PR #42, opened 2025-10-08
Mixer Effects

tobyspark:
...aka kicking the tyres

### DIFF
--- a/Fabric.xcodeproj/project.pbxproj
+++ b/Fabric.xcodeproj/project.pbxproj
@@ -95,6 +95,7 @@
 			isa = PBXFileSystemSynchronizedBuildFileExceptionSet;
 			membershipExceptions = (
 				Effects,
+				EffectsThreeChannel,
 				EffectsTwoChannel,
 				Extensions/CGAffineTransform.swift,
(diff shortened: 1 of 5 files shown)


vade:
Nice!

vade:
This looks good to me! Only thought i might have, since this is general purpose, is there a more user legible name for the SPK MXR (which i admit has a cool ring to it haha) 

exited to have your contributions!

tobyspark:
The naming issue really stems from the fact it should be a plug-in node that I make available. Then SPK xxxxx makes more sense. For now, we suck it up as the price of contribution? Having node descriptions would help too, but that ain’t MVP.

vade:
Is the expectation this ships with the main editor / framework, or is this a temporary try?  Maybe the thing to do is include the 3 way mixer base class which i was intending to do?

tobyspark:
Dunno. I'm not too fussed about what ships in an alpha? Especially in the absence of some kind of plugin loading system?

vade:
Ok, my concern mostly stems from some legibility concerns (which, granted need to be sorted with better in app docs). 

We'll figure it out!

PR #62, opened 2025-10-29
Add a draft base compute node

vade:
Working on Satin - Fabric compute support with some Base node implementations.


vade:
Goal here is to be able to run certain kinds of algorithms on GPU, like MarchingSquares or Contour algorithms to produce viable Geometry from masked images.

vade:
This is far enough along ill just merge to dev YOLO etc etc

PR #71, opened 2025-11-09
App Icon

tobyspark:
via Icon Composer. I’ve changed the app icon name in the target settings, but don’t see the icon being picked up. 

https://developer.apple.com/documentation/xcode/creating-your-app-icon-using-icon-composer/#Add-your-Icon-Composer-file-to-an-Xcode-project


vade:
Thanks for this! Apologies for the tardy reply, been traveling and haven't had a chance to catch my breath!

PR #75, opened 2025-11-15
Fix modulo by zero crash

JustinMeans:
Adds zero check to modulo operator to prevent crash when divisor is 0.

Returns 0 for modulo by zero, matching existing division operator behavior.


vade:
Thank you!

PR #76, opened 2025-11-15
Add bidirectional node connection dragging

JustinMeans:
Enables dragging node connections in both directions (outlet→inlet and inlet→outlet).

Previously only outlet→inlet dragging was supported. This adds inlet→outlet capability by mirroring the drag/drop pattern for inlets.


vade:
Ah awesome. Appreciate that. It’s funny this didn’t even occur to me as I never drag from outlets. Much obliged!

PR #77, opened 2025-11-15
Add drag preview line

JustinMeans:
Shows dashed preview line while dragging node connections.

Fixes curve rendering when dragging right-to-left.

- DragGesture tracks cursor position during drag
- Fixed UTType naming (hyphenated identifiers for macOS compatibility)
- Direction detection for proper curve rendering

https://github.com/user-attachments/assets/d2d92cca-09ca-4e99-bda0-486a4799fa6a

### DIFF
--- a/Fabric Editor/Info.plist
+++ b/Fabric Editor/Info.plist
@@ -28,24 +28,16 @@
 			<array>
 				<string>public.data</string>
 			</array>
-			<key>UTTypeIcons</key>
-			<dict/>
 			<key>UTTypeIdentifier</key>
-			<string>info.vade.v.outletData</string>
-			<key>UTTypeTagSpecification</key>
-			<dict/>
+			<string>info.vade.v.outlet-data</string>
 		</dict>
 		<dict>
 			<key>UTTypeConformsTo</key>
 			<array>
 				<string>public.data</string>
 			</array>
-			<key>UTTypeIcons</key>
-			<dict/>
 			<key>UTTypeIdentifier</key>
-			<string>info.vade.v.inletData</string>
-			<key>UTTypeTagSpecification</key>
-			<dict/>
+			<string>info.vade.v.inlet-data</string>
 		</dict>
 	</array>
 	<key>UTImportedTypeDeclarations</key>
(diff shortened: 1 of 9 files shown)


vade:
Thanks for this! I’ve been meaning to get to this feature. I’ve yet to have a coffee and want to do a proper review but I wanted to share my appreciation first. 

Cheers for the PRs !!

JustinMeans:
Thanks for the awesome project! I've been messing with it the past few hours. This PR needed a quick fix but in my testing has been a good upgrade and should be ready to go.

vade:
Awesome! It’s def gonna be a nice improvement to the UI/UX experience!

vade:
One minor side effect of the work here appears that node inlets / outlets can have a negative (inward to node rect) facing coord stem:

image

JustinMeans:
Good catch, that was some conflicting logic with the new inlet -> outlet in #76. This commit should fix that.

vade:
Hrm, closer but this feels wrong to me:

image


Seems like the furst curve post stem has the wrong direction on the control point or some such? 

What was the reasoning to mess with the path drawing? I think it was ok for this case?

vade:
And to be clear, thanks again for the slew of PRs! Its really nice to see activity on the project!

JustinMeans:
I agree that looks off... admittedly, the first two PR's which merged were manual but then I spun up Claude and it went a bit overboard on the next 2 PRs. Have to step away for a bit but will try to tackle more later this weekend. Excited about the project!

vade:
Haha all good. I appreciate the enthusiasm!

I suspect resetting the drawing code in `calcPaths` likely would get very close to the visually correct drawing along with line preview which is quite nice!

Thanks again!

BTW I checked out some of your work, it seems you have some experience with Stream Diffusion like realtime video gen on Apple Silicon? 

I was pondering seeing if a CoreML / ANE powered stream diffusion like thing could be done in Fabric via MLX / CoreML and some caching of intermediate layer output to safe on compute. 

Anyway great to have some other eyes on this!

JustinMeans:
Sweet, feel free to make the changes and close here if needed, I'm easy on attribution etc. Just excited to see and use a project like this and hope it accelerates fast. I'll jump back in on more development on the repo in a day or so. 

Much of my interest is in developing fast image / video on  in MLX and CoreML indeed. I think the combination with a tool like this would be awesome! I'm tinkering with MLX-Swift for realtime generative video and can envision how they come together. It's still primitive but would love to make it happen!

PR #78, opened 2025-11-15
Fix scene not updating when connections change

JustinMeans:
Metal render view doesn't update when node connections are added/removed.

Still fixing and finding better solution, WIP.

### DIFF
--- a/Fabric/Graph/Graph.swift
+++ b/Fabric/Graph/Graph.swift
@@ -507,15 +507,12 @@ internal import AnyCodable
     
     public func syncNodesToScene(removingObject:Object? = nil)
     {
+        self.scene.removeAll()
+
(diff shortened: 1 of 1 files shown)


vade:
Ah, that tracks! Lovely fix! Totally missed this cause. Nice catch.

vade:
Actually, pondering this fix / bug, i suspect its also responsible for memory retain cycle in the graph code. 

Nice.

JustinMeans:
Cool, feel free to make whatever changes you want. I wasn't sure if this was best performance-wise so changed to a draft.

vade:
Yea, i was / am leery of doing too much scene graph fiddling but correctness over perf (generally) at this stage of the game!

PR #82, opened 2025-11-16
Make Xcode project generic for any developer

JustinMeans:
Removes hardcoded team ID and uses Apple's sample code bundle identifier pattern to allow any developer to clone and build.


vade:
hey Justin, thanks for yet another PR!

Can you help me understand the motivation here? Im honestly not sure about this one! 

My thinking here is  we want to get Fabric.framework and Fabric Editor to have  canonical bundle ID's - since the SDK's goal is to eventually be embeddable and Fabric doc's as interchange format, i think its helpful to have a stable and consistent bundle ID for various reasons. 

Is the idea here that the team ID is just a pain to change for building oneself locally? 

Maybe theres a spec of design pattern im unaware of for the example code bundle ID ? If so, is there a doc or guiding principle here? 

Thanks!

JustinMeans:
`Is the idea here that the team ID is just a pain to change for building oneself locally?`
That was the idea, helpful for cloud agents and setup on each pull, but makes sense to have the bundle ID for frameworks. Revised to just remove hardcoded team id and keep the bundle.

vade:
ah cool cool. Thanks i appreciate these quality of life fixes!

Issue #66, opened 2025-10-31
SwiftUI Optimizations


vade:
some insight on environment usage here

https://www.youtube.com/watch?v=yXAQTIKR8fk&t=5102s

vade:
Ok, so after having pulled my hair out getting instruments profiling to work, I see that

* node registry views binding to scrollOffset causes a lot of view slowdown - WAY more than we would expect.

We need to find a nice way to get scroll offset cached somewhere appropriate and not inside of our view.

vade:
PortAnchorKey should also be looked at.

PR #83, opened 2025-11-16
Add scroll-to-node and zoom features

JustinMeans:
1: Adds pinch-to-zoom (two-finger magnify gesture)

https://github.com/user-attachments/assets/10089b82-c398-4c45-9d7f-8389b42b1a64

The SwiftUI Canvas + ScrollView is notorious to get the pinch magnification right but this is a first-step and adds value where it's at. 

2: Scrolls to first node when opening a project.

Problem: When opening a project, it doesn't scroll to the first node. Test it with this example:
[Test Sphere.zip](https://github.com/user-attachments/files/23565597/Test.Sphere.zip)
Before: Project opens in top left corner, lost in the canvas, have to scroll all the way to center to find nodes.
After: The ScrollView automatically scrolls to the nodes

### DIFF
--- a/Fabric Editor/Views/ContentView.swift
+++ b/Fabric Editor/Views/ContentView.swift
@@ -8,7 +8,10 @@
 import SwiftUI
 import Fabric
 
-
+struct ScrollGeometry: Equatable {
+    let offset: CGPoint
+    let containerSize: CGSize
+}
 
 struct ContentView: View {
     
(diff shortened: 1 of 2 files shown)


vade:
nice! I had a version of pinch to zoom working, but didnt get it to a place where patching / wireing nodes up worked while zoomed out. Does your work solve the coordinate transforms for that to work? 

Nice stuff!

JustinMeans:
It does solve the coordinate transforms by only applying the scale effect to the root view. Of course when you're _too_ far zoomed out you won't hit the hitbox. Went through a couple iterations before that and had the same issue, but this version seems to have solved it in my testing.

https://github.com/user-attachments/assets/537229b1-f375-4b53-94ce-c95d14dd5e3c

JustinMeans:
It does have that offset / stutter though when the magnification gesture initiates, as seen in the recording... been wrestling with that and SwiftUI but haven't gotten it better than this. Still a starting point in the right direction and nice to have building out larger files.

vade:
One of the stutter/ scroll issues is due to use of @\environment, which ive been meaning to factor out. Apparently @\environment cascades recalculations through the view tree, stopping some optimization from happening that the new @\Observable mechanisms support. 

https://github.com/Fabric-Project/Fabric/issues/66 has a link to a newer Apple video espousing some advice for optimizing some view redraw logic. 

Nice stuff with the coordinate transform just working!!

JustinMeans:
Oh interesting, good to know. I'll check that out 👍🏻

PR #84, opened 2025-11-16
Fix zoom jump by anchoring magnification to viewport center

JustinMeans:
Addresses feedback from #83 regarding zoom stutter caused by @Environment observation cascades. 

**Changes:**
- Calculates zoom anchor from current scroll position instead of gesture location to eliminate jumps between magnification gestures

**Result:** Smooth zoom to viewport center wherever scrolled, no jump on gesture initiation/termination.

https://github.com/user-attachments/assets/015e2a8d-cca1-4a50-bd1f-f90bc4c01515

Related: #83, #66

### DIFF
--- a/Fabric Editor/Views/ContentView.swift
+++ b/Fabric Editor/Views/ContentView.swift
@@ -8,54 +8,45 @@
 import SwiftUI
 import Fabric
 
-struct ScrollGeometry: Equatable {
-    let offset: CGPoint
-    let containerSize: CGSize
-}
-
 struct ContentView: View {
-    
+
     @Binding var document: FabricDocument
-    
-    @GestureState private var magnifyBy = 1.0
-    
+
+    @State private var magnifyBy = 1.0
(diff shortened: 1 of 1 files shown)


vade:
Thanks! Post merge of the last PR I caught the old deprecated magnify gesture - I was working on similar changes but you beat me to it :) 

At a glance this looks good. Will review later tonight! Again thank you!

vade:
So this method seems to cause `Modifying state during view update, this will cause undefined behavior.` due to view change callbacks in `onScrollGeometryChange` updating state in the view itself

Also this PR introduces visual glitches when scrolling with certain zoom levels, likely due to incoherent values due to state updates during view callbacks. 

ill poke around, because overall the zoom anchor behaviour def feels right, but the overall resulting behaviour could use some work. 

Thank you again for the PRs!

vade:
Did a smidge of cleanup on my own branch of this - Thanks again, these PRs are awesome! Keep em coming!

PR #85, opened 2025-11-16
Add UndoManager support for graph operations

JustinMeans:
Implements undo/redo for node add/delete and port connect/disconnect.
Fixes connections failing to update in render view during certain operations.


vade:
Thanks!

a brief review on mobile shows me some dispatch async code in the registration of the undo manager. Can you walk me through the rational for that? Anytime I see async code that also handles the main queue it’s a moment for me to pause and consider!

PR #96, opened 2025-11-18
Feature/add additional geom nodes

vade:
Add Arc, Cone, Tube, Torus Geometry Nodes


jvcleave:
if you are getting compile errors - you need to add these to the target

image

vade:
this is fixed in main

PR #113, opened 2025-11-23
Collision-check and cascade added nodes. Fixes #107

### LINKED ISSUE #107: Adding multiple patches should clearly indicate their addition to the graph
### Issue

Clicking a patch in the library more than once ends up creating multiple patches, but you can't tell because the patches are created instantly and they are on top of each other

### Proposal

Options:
1. Drop the patches onto the grid with a bounce
2. Offset them diagonally down and to the right

tobyspark:
+ Visibility check on add
+ Scroll-to functionality
+ 'Center on selected node' menu item
+ DRY node anchor point refactor

### DIFF
--- a/Fabric Editor/FabricApp.swift
+++ b/Fabric Editor/FabricApp.swift
@@ -45,6 +45,7 @@ struct FabricApp: App {
             {
                 CheckForUpdatesView(updater: updaterController.updater)
             }
+            ViewCommands()
         }
         
         Window("About Fabric Editor", id: "about") {
@@ -65,5 +66,31 @@ struct AboutCommands: Commands {
                 openWindow(id: "about")
             }
         }
+    }
+}
+
+struct ViewCommands: Commands {
+    @FocusedValue(\.centerOnSelectedNode) private var centerOnSelectedNode: (() -> Void)?
+    
+    var body: some Commands {
+        CommandGroup(after: .sidebar) {
+            Button("Center on Selected Node") {
+                centerOnSelectedNode?()
+            }
+            .keyboardShortcut("f", modifiers: .command)
+            .disabled(centerOnSelectedNode == nil)
+        }
+    }
+}
+
+// FocusedValue key for center on selected node action
+struct CenterOnSelectedNodeKey: FocusedValueKey {
+    typealias Value = () -> Void
+}
+
+extension FocusedValues {
+    var centerOnSelectedNode: CenterOnSelectedNodeKey.Value? {
+        get { self[CenterOnSelectedNodeKey.self] }
+        set { self[CenterOnSelectedNodeKey.self] = newValue }
     }
 }
--- a/Fabric/Views/NodeRegisitryView.swift
+++ b/Fabric/Views/NodeRegisitryView.swift
@@ -12,13 +12,20 @@ public struct NodeRegisitryView: View {
 
     public let graph: Graph
     @Binding public var scrollOffset: CGPoint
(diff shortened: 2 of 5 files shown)


vade:
Minor nit's incoming - apologies!

Fabric/Views/NodeRegisitryView.swift
```diff
@@ -120,4 +134,41 @@ public struct NodeRegisitryView: View {
         let nodesForType:[NodeClassWrapper] = availableNodes.filter( { $0.nodeType == nodeType })
         return  self.searchString.isEmpty ? nodesForType : nodesForType.filter {  $0.nodeName.localizedCaseInsensitiveContains(self.searchString) }
     }
+    
+    /// Checks if a node's bounds are visible in the current viewport
+    /// Returns true only if the entire node is within the visible area
+    private func isNodeVisible(center: CGPoint, size: CGSize) -> Bool
```
vade: This func feels like it belongs on `Node` model, not in nodeRegistry, as it might be helpful elsewhere, or in a layout manager class which encapsulates some of the layout calcs used 'across' views.

vade:
Thanks for this PR

My feeling on this 

* We might want a graph layout manager class?
* Layout manager class can have some functions for a proposed node offset given some input / visibility check
* We can maybe refactor some of the other views to have simpler data models / bindings and clearer separation of responsibilities

My concern here is there is no longer a centralized spot for layout decisions, and by having disparate objects / views handle it makes it hard to understand the ownership of decisions

I vote for separation of responsibilities - 

Sorry if this feels like yucking the work, its not my intention, and im responsible for some of the above already!

Fabric Editor/FabricApp.swift
```diff
@@ -67,3 +68,29 @@ struct AboutCommands: Commands {
...
+        }
+    }
+}
+
+// FocusedValue key for center on selected node action
+struct CenterOnSelectedNodeKey: FocusedValueKey {
+    typealias Value = () -> Void
+}
+
+extension FocusedValues {
```
vade: Is this strictly necessary? I think the focus value is for keyboard/device input, which in the canvas shouldnt be a thing unless youre interacting with the node via input device somehow? 

Do we need to track focus as well as selection? If so, cant we handle the binding using the .focus on the canvas view ? 

I ask because a node can be selected, but not focused (focus might be in the inspector side panel, or the node registry for example?)

tobyspark:
Ya. I think the PR as-is is as well factored as is reasonable given what was there already. If we want to do this right, then yes I think laying-graphs-out and managing-a-graph-to-traverse-and-execute are two quite different things. First question would be – is the layout entity entirely responsible for layout, e.g. graph node entities don’t have position properties, or is more a  extension of the graph. Could even talk this through on a call.

tobyspark:
> First question would be...

Let’s pick this up.

vade:
Hey hey!

I wasnt sure if you wanted to respond to my inline comments / review? I had some pointed questions about the implementation details that i dont think have been addressed. LMK

vade:
Hey - i replaced some of this functionality with a pretty light weight surgical fix only to graph, which adds a simple timeout heuristic at 1 second for consecutive node ads to help with clarity when quickly adding things. 

if that isnt sufficient, or we want it always on (ie always offset no matter the time delay) we can remove the check, or have it reset in a different manner (scroll, move node or some other hueuristic), but i think this solution touches far less code and keeps the graph solely responsible for layout for now without introducing any other stuff

PR #127, opened 2025-11-27
Feature/syphon

vade:
PR Adds Syphon Support

One unexpected nuance - since Fabric supports Half Float textures, and Syphon supports BGRA8Unorm, we seem to have a color issue stemming from gamma perhaps? :X 

- [x] Server
- [x] Client
- [ ] Server List
- [ ] Color Correction


vade:
Given how Syphon Server directory is greedy and returns the first server even if there are no exact matching strings i think we can avoid the server list for now ?

vade:
For color, i need to update Lygia metal tone mapping support most likely.

PR #142, opened 2026-01-09
Rename node functionality

### LINKED ISSUE #123: Now that subgraphs are working well, we should be able to rename nodes and have that save
I think we have the parts for this on the back end, but we need a nice UI to enable renaming of nodes ideally (and maybe even ports eventually)

furtunavlad:
Closes #123
Steps to test:
- In canvas, double-click a node or right-click->rename

### DIFF
--- a/Fabric/Views/Nodes/NodeView.swift
+++ b/Fabric/Views/Nodes/NodeView.swift
@@ -69,9 +69,9 @@ struct NodeView : View
     //                        .foregroundStyle(.primary)
                             .foregroundStyle(  self.node.nodeType.color()  )
                             .frame(maxHeight: 20)
+                            .contentShape(Rectangle())
                             .padding(.top, 5)
                             .padding(.horizontal, 20)
-                            .contentShape(Rectangle())
                             .onTapGesture(count: 2) { // double click to rename
                                 renaming = true
                             }


vade:
Thanks so much! This looks good on first pass  eyeball review! Gimme a bit to check it out in depth!

Thank you again!

vade:
So one thing I would say, is we might want to limit double click to rename, as that 'enters' a subgraph node (nodes which have a more square body like Subgraph, Environment, Render to Image, Iterator) - if this usurps the double click for the node, it might be problematic!

Fabric/Views/Nodes/NodeView.swift
```diff
@@ -36,14 +40,42 @@ struct NodeView : View
...
+                        Text( self.node.name )
+                            .font(.system(size: 9))
+                            .bold()
+    //                        .foregroundStyle(.primary)
+                            .foregroundStyle(  self.node.nodeType.color()  )
+                            .frame(maxHeight: 20)
+                            .padding(.top, 5)
+                            .padding(.horizontal, 20)
+                            .contentShape(Rectangle())
+                            .onTapGesture(count: 2) { // double click to rename
```
vade: I see - we might be able to get away with this, as its double click on the title vs the body of the node. 

My thinking here, as the view is getting large  - maybe we pull out to a new view `NodeTitleView` which doesnt have styling defined in it, but just a group which wraps the if / else and conditionally has the editing or not, and has a binding to trigger the editing?

furtunavlad:
updated to a smaller hitbox for double click to rename

vade:
Thanks! I made a minor refactor to extract the logic to a self contained view, and remove some extraneous binding i didnt see was necessary! This is really nice! Thank you!

PR #161, opened 2026-01-21
#158 - Add Mouse Scroll Wheel Zoom support in canvas

### LINKED ISSUE #158: Support zooming of node graph with modifier key and mouse scroll wheel
It would be great if you could hold down a modifier key, and then scroll the mouse scroll wheel to change the zoom of the node graph.

latenitefilms:
- Fixes #158

### DIFF
--- a/Fabric Editor/Views/ScrollWheelZoomEventMonitor.swift
+++ b/Fabric Editor/Views/ScrollWheelZoomEventMonitor.swift
@@ -22,6 +22,7 @@ struct ScrollWheelZoomModifier: ViewModifier {
                     consumeScrollEvent: consumeScrollEvent,
                     onZoom: onZoom
                 )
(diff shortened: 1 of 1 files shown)


vade:
Hey! Some high level thoughts here

We might want this to be a view modifier, rather than an overlay? 

I did some looking at this, and saw  folks use something like below which might be less burdensome?


Since a view is a struct, i dont think we get any memory issues one shotting the addLocalMonitor ?
```
 .onAppear(perform:{
                NSEvent.addLocalMonitorForEvents(matching: .scrollWheel) {event in

                // handle event and dont consume
                    return event
                }
            })
 ```

vade:
Also thanks for taking the time to look at solves for this!

latenitefilms:
@vade - Let me know if this is any better? Obviously more than happy for you to make any tweaks or changes you need to!

PR #170, opened 2026-02-07
WIP Plugin system for Fabric, mirror QCPllugin style api

vade:


vade:
We need to add - for plugins which expose or provide metal shaders a way to know which metal file to load, or ideally precompiled metal library to load 

this means re-working some assumptions in the plugin loading, and codable implementation (a node should likely track the source plugin / plugin id or some such it loads from?)

tobyspark:
Superseded by #303

PR #179, opened 2026-03-01
TestCardProviderNode

tobyspark:
A very basic test card image provider, implemented as a compute shader for pixel-precision.

### DIFF
--- a/Fabric/Compute/TestCard.metal
+++ b/Fabric/Compute/TestCard.metal
@@ -19,36 +19,61 @@ kernel void testCardGenerate
 
     if (gid.x >= w || gid.y >= h) return;
 
-    bool onLine = false;
-
-    // Border: top, bottom, left, right edges
+    // 1px white border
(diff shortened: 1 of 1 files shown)


vade:
you might be interested in the Lygia shaders shapes / drawing / sdf functions which can run in compute too

tobyspark:
I tried that and didn’t get the exact results I expected. Tho’, it was somewhat cursory, and regrettably I didn’t take a screenshot. I need a fancier version of this node myself so will have another dip into this soon. In the meantime... you’ll see the next PR, which is related.

vade:
OK, so i dont think its a huge issue for this PR, but Satin provides some existing compute infrastructure, and I set up `BaseTextureComputeProcessorNode` 

I'll merge this as is, but i might refactor to have a `BaseTextureComputeProviderNode` and make the Test Card a subclass therein if you dont mind:)

PR #180, opened 2026-03-01
BillboardNode

tobyspark:
It’s quite a faff hooking up ‘Image Material’, ‘Image Dimensions’, ‘Plane Geometry’, and ‘Mesh’ just to get an image on-screen. As per QC, this is a single node to get an image input on-screen.

Two issues
- Setting width=2 and then resizing the output window does not have the expected effect, e.g. it scales with the rendering destination width.
- ~~This makes me think about coordinate systems. The perspective and ortho cameras appear to have different systems (perspective is width -1,+1; orthogonal is height -1, +1). Having a setting for which axis is the sizing dimension would be very useful on each camera type. Plus the perspective camera needs FOV? I’ll start discussion / work for this elsewhere, but not today, so flagging here given context.~~ PR Incoming

### DIFF
--- a/Fabric/Nodes/NodeRegistry.swift
+++ b/Fabric/Nodes/NodeRegistry.swift
@@ -56,6 +56,7 @@ public class NodeRegistry {
         InstancedMeshNode.self,
         InstancedModelMeshNode.self,
         EnvironmentSkyboxNode.self,
+        BillboardNode.self,
     ]
     
     private var geometryNodeClasses: [Node.Type] = [ // Geometry
--- a/Fabric/Nodes/Object/Renderables/BillboardNode.swift
+++ b/Fabric/Nodes/Object/Renderables/BillboardNode.swift
@@ -1,0 +1,104 @@
+//
+//  BillboardNode.swift
+//  Fabric
+//
+//  Created by Toby Harris + Claude Opus 4.6 on 3/1/26.
+//
+
+import Foundation
+import Satin
+import simd
+import Metal
+
+class BillboardNode: BaseRenderableNode<Mesh>
+{
+    override public class var name: String { "Billboard" }
+    override public class var nodeType: Node.NodeType { .Object(objectType: .Mesh) }
+    override public class var nodeExecutionMode: Node.ExecutionMode { .Consumer }
+    override public class var nodeTimeMode: Node.TimeMode { .None }
+    override public class var nodeDescription: String { "Renders an image on a plane with automatic aspect-ratio sizing" }
+
+    override public class func registerPorts(context: Context) -> [(name: String, port: Port)] {
+        let ports = super.registerPorts(context: context)
+
+        return [
+            ("inputImage", NodePort<FabricImage>(name: "Image", kind: .Inlet, description: "Image to display on the billboard")),
+            ("inputSize", ParameterPort(parameter: FloatParameter("Size", 1.0, .inputfield, "Size of the billboard in world units"))),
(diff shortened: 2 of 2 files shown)


tobyspark:
Oh, also, we could call this `Sprite` instead of `Billboard`. I don’t think we need QC's two, and the base provider class already gives some ‘sprite-ness’, so could standardise on the more obvious _sprite_.

vade:
Thanks for all the PRs! 

Quick thoughts: 

So I sort of went back and forth on this approach (single nodes vs keeping mesh + geo + material), and im a fan of consistency. While i grant you 100% that having separate geo and materials is right now onerous (until we get clipping or what not) it keeps a consistent model of how things should work. 

In my mind, billboarding is a specific material effect on how perspective is applied in the vertex shader on the material, and any shape could get billboarded in theory. 

Here are the user concerns  i see that i think we should find a holistic and standard path for

- I want stuff to always draw on screen full size aspect correct regardless of camera position and coordinate system 
    - making it easy to mix 3D with image video background
- I want stuff to always face the camera regardless of the camera orientation and position 
    - particles, effects, sprite sheet characters,
- I want to images / content graphically in space

QC Sprite -> Plane Geo + Whatever Material + Mesh
QC Billboard -> Plane Geo + Specific billboard material option enabled, or use a specific material + Mesh

Questions for you:

* how does this work with lighting? 
* how does this work with shadows?
* how does this work with blending? 
* do we need to compose some material settings into this node for control?
* Is part of this kind of a reaction to the complexity / annoyance of using mesh + geo + material to wire shit up?

I can see a place where a Billboard node makes it clear

* this is screen aligned quads
* No shadows recieving
* No shadow casting
* No lighting
* Limited blend mode (none, alpha, additive hardware blending)

But im not yet convinced. 

Thanks for this, im not trying to be a stickler, i just want to strive towards a well composed, fully thought out, ideally everything works together system, and the choice of separating out materials from geometry makes some things more tedious to set up, but ultimately (maybe?) more flexible and composable. 

LMK your thoughts.

tobyspark:
The graph needed to go from previewing an image in the Finder to seeing that in the Fabric output window is _a lot_. It’s arguably project-killing-complex, as in too user-unfriendly. Even if you get the nodes, the understanding needed to fill the output, aspect correct is a lot, and hidden (e.g. not just the maths, but our coordinate system etc.). 

So I 100%, 1000% think Fabric needs something like a Sprite/Billboard node. It’s the first thing a lot of people will do, so we don’t want them to bounce off, and it’s something I used the hell out of for real-world production work in QC. 

To your specifics, I think the basic answer is “if there’s an option, the whole point of this is to remove the option; pick the simpler”. But also, don’t a lot of those options come with the base node? So we’d actually have to work to hide them? I haven’t used what I have here in practice yet due to a) the sizing issue and b) I’ve run out of time today, but I don’t have any particular flags in my head that it’s too much or too little.

tobyspark:
Postscript – I’m arguing there needs to be a _Sprite-like_ node there for a) beginners and b) everyday convenience. Whether that is an actual node, as here, or some kind of macro-patch, I’m open to. But if the latter, that’s a whole new feature (and quite a large design space and UX to get right).

vade:
So thinking a bit about this, we might want to call this Image Mesh (?) , as Billboard is sort of specific QC parlance that itself might be confusing. I get what you mean about all the caveats.

we should roadmap 

1) plan for coordinate system handling
2) actual billboarding material that is front facing to camera at all times. 

will merge as is, and we can workshop names and shit later. 

Thanks!

Issue #182, opened 2026-03-01
provide Graph api for getting list of all published inputs and output ports


vade:
Just a reminder to expose some stuff for @tobyspark

PR #183, opened 2026-03-02
Clean up port API on Graph, exposing published ports  to external fra…

vade:
…mework clients and keeping all ports internal so folks can poke around too much.


vade:
fix #182

PR #191, opened 2026-03-14
Feat – Engine gets 'active node ports’, enabling performant 'Switch' node

tobyspark:
A node to switch between various inputs, per an index input.

Ideally this should “turn on and off” the different branches of the graph connected to the switch node. The GraphRenderer previously would traverse all input ports of a node, this PR adds the concept of ‘active ports’ to make this possible.

### DIFF
--- a/Fabric/Graph/Graph.swift
+++ b/Fabric/Graph/Graph.swift
@@ -67,6 +67,9 @@ internal import AnyCodable
     // Fix for #103 - connection/topology changes trigger syncNodesToScene() inside of `GraphRenderer`.
     @ObservationIgnored private var pendingConnectionSceneSync = false
     public private(set) var connectionRevision = 0
+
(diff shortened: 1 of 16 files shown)


tobyspark:
Marked as draft. This will definitely need some more testing and likely some discussion. In QC-speak, this node is ‘multiplexer’, and that might yet be a better name. Switch is more discoverable, but doesn’t pair with the opposite ‘demultiplexer’.

FWIW, I made this along the way doing my app, but ended up not needing it as I hardcoded the logic I needed this node for elsewhere. Used multiplexers all the time in QC though.

tobyspark:
I think _Gate_ is a good name to pair with _Switch_. All I’ve got as alternative is _Switch In_ and _Switch Out_. Or QC’s _Demultiplexer_ and _Multiplexer_ but I think that’s not discoverable and the way I used them at least was not in the spirit of muxing.

PR #194, opened 2026-03-14
Add TextureCropNode – GPU blit crop for texture sub-regions

tobyspark:
Crops a rectangular region from an input texture via MTLBlitCommandEncoder, producing a smaller output texture. Uses blit rather than UV-based sampling because Fabric currently lacks UV offset/scale machinery in its material and geometry nodes (BasicTextureMaterial shader samples geometry UVs directly, PerspectiveQuadGeometryNode bakes UVs as 0→1).

The blit approach is performant for typical use: one cheap GPU copy per slice, and the smaller downstream texture improves cache locality and reduces bandwidth for all meshes that sample it.

### DIFF
--- a/Fabric/Nodes/Image/TextureCropNode.swift
+++ b/Fabric/Nodes/Image/TextureCropNode.swift
@@ -1,0 +1,64 @@
+import Foundation
+import Satin
+import simd
+import Metal
+
+/// Crops a rectangular sub-region from an input texture using a blit encoder.
+public class TextureCropNode: Node
+{
+    public override class var name: String { "Texture Crop" }
+    public override class var nodeType: Node.NodeType { .Image(imageType: .BaseEffect) }
+    override public class var nodeExecutionMode: Node.ExecutionMode { .Processor }
+    override public class var nodeTimeMode: Node.TimeMode { .None }
+    override public class var nodeDescription: String { "Crops a rectangular region from a texture" }
+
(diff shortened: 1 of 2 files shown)


tobyspark:
I have a specific use for this in my app. Do we want it in Fabric generally? 

(Fabric doesn’t have a way to tell downstream nodes to only work on a region of a texture, effectively setting UV? QC did... you used a Image Texturing Dimensions or somesuch node?)

Also the commit comment here might not apply generically, was written in the context of my app.

vade:
Yea, so QC leveraged QCImage / CIImage to handle region of interest shit. Its a huge undertaking and ive avoided it for a reason. 

1) all pro apps at apple have migrated from Core Image to use their own metal implementation for speed, CI is kind of.. not as popular for realtime shit these days (slow, cumbersome and lot of overhead)

2) most use cases ive had i dont need it. Im not doing any domain of definition / region if interest work and almost all effects work 'full faster' that ive made, needed or seen in the wild, save for times where you want specific kinds of behaviour. 

I think a node like this is perfect due to the above.

vade:
Oh, one thing we could do is have a texture matrix associated with the Fabric Image, allowing texture rotation, offset, scaling for free, but Satin doesnt have that in its material pipeline either. So we should not implement till Satin has some work there on its default material pipeline to support it. 

QC required a texture matrix multiply to compute the proper UV coords in the vertex shader for fragment processing. Satin just doesnt have that spread around yet.

PR #196, opened 2026-03-14
Fix negative deltaTime from CAMetalDisplayLink

tobyspark:
CAMetalDisplayLink's callback deltaTime parameter is not elapsed time since last frame — it's negative. Replace with our own monotonic delta computed from consecutive CACurrentMediaTime() calls.


vade:
Nice catch!

PR #205, opened 2026-03-15
Feat – Tween nodes

tobyspark:
The _Number Ease_ node provides the easing curves, but doesn’t actually animate/tween/smooth a number. The _Number Smooth_ does smooth a number, but is a specific algorithm.

This provides two nodes that use the easing functions of _Number Ease_ to smooth incoming values over time. One for numbers, and one for colours that takes the vec4 and performs the tween in OkLab colourspace.

### DIFF
--- a/Fabric/Nodes/NodeRegistry.swift
+++ b/Fabric/Nodes/NodeRegistry.swift
@@ -245,6 +245,7 @@ public class NodeRegistry {
         MathExpressionNode.self,
         GradientNoiseNode.self,
         AudioSpectrumNode.self,
+        ColorTweenNode.self,
         ArrayFirstValueNode<Float>.self,
         ArrayLastValueNode<Float>.self,
         ArrayIndexValueNode<Float>.self,
--- a/Fabric/Nodes/Parameters/Color/ColorTweenNode.swift
+++ b/Fabric/Nodes/Parameters/Color/ColorTweenNode.swift
@@ -1,0 +1,154 @@
+//
+//  ColorTweenNode.swift
+//  Fabric
+//
+
+import Foundation
+import Satin
+import simd
+import Metal
+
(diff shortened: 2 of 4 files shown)


tobyspark:
Thoughts
- should there be a single ‘Smooth’ node that can handle any smoothable type?
- should this have the option of blended tweening rather than the simpler snap-and-retarget implemented here?
- should this also have sprung smoothing as easing option or separate node?

vade:
are all colors assumed to be linear gamma? Im trying to make Fabric an unbiased renderer where blending happens in linear space before being corrected, so we dont need to add debiasing on every input and add it on output again (saves cycles but requires assumptions to be stuck to)

vade:
Answered my own q - thanks for this looks good to me!

tobyspark:
I don’t think there’s a written policy for RGB values “in the graph” vs. internal processing? At least I asked Claude to audit this and it didn’t come up with much.

We should have something written to be consistent about before we merge this.

vade:
Yea, i have notes in some markdown files somewhere about color handling in Fabric with a goal for linearity, but i agree the 'contracts' are not super clear or definitive yet. Overall

* all colors, including image colors and render target colors should hit linear gamma wide color / HDR - im tempted to target ACES linear but its a big challenge. 

This means

* image loading should be aware of and conform images to linear space
* video as well 
* generators
* single value cpu side color / gradient or whatever generators and converters
* outputs will be responsible for mapping linear wide gamut -> display or target output.

Generally this is a scene referred pipeline with input color spaces expected to confirm to a shared working space, which then is conformed per output to the destination space.

tobyspark:
That should be in ARCHITECTURE.md ?

Note there’s also a prospective glossary now in https://github.com/Fabric-Project/Fabric/pull/207, more for the UX side.

PR #207, opened 2026-03-15
Add project glossary with terminology

tobyspark:
...a starting point for discussion, via the Genie.

> Catalogues all domain terms from the codebase, documents six terminology inconsistencies (Producer/Provider, Patch/Node, Geometery typo, Inlet-Outlet/input-output, Image/Texture, Object overloading), and identifies eight unnamed concepts (value propagation, dirty cascade, port interface, channel count, execution lifecycle, scene proxy, default camera, managed/unmanaged image lifetime).

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -1,0 +1,217 @@
+# Fabric Glossary
+
+A reference for Fabric's domain terminology, covering core concepts, known inconsistencies, and concepts that would benefit from naming.
+
+## Core Concepts
+
+### Graph
+A complete Fabric document: the set of all Nodes and their Connections. A Graph is serialisable (`Codable`), versioned, and supports undo. When saved to disk, a Graph is sometimes referred to as a **Composition** (Quartz Composer heritage), though the codebase uses `Graph` exclusively.
+
+### Node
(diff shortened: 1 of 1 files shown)


tobyspark:
This is missing ‘Value Nodes’, the possible term for what QC called an input splitter?

tobyspark:
...or are those _Parameter Nodes_ ?

vade:
Ive been calling them parameter nodes, as per NodeType, which does encode some of this in code.

vade:
So my only concern is i think there are some poor design decisions in the graph side of things (GraphExecutionContext design being one of them). As long as we remember to sync docs when refactoring things.

tobyspark:
I thought we might do some passes in this PR, but I guess I didn’t mark it as draft! Anyhow, it’s in now and so... lets pick this up on a regular* call.

On keeping things in sync – this is one thing I think genies are an unequivocal win for. They see what’s actually there not what you remember, and can go through it all. There’s some kind of world where a CI action compares a fresh analysis with this the architecture doc.

And – I’ll update the “insert input splitter” PR to _Parameter Nodes_.

\* regular but not necessarily frequent

Issue #198, opened 2026-03-14
Drag and drop support from Finder to Canvas


vade:
We want to allow for any nodes which support file loading via the File Loading node protocol extension to support auto instantiation of a new node with a populated file url path.

We should

* Extend `NodeFileLoadingProtocol` to expose canonically supported mime types or UTI arrays as a static class proprty
* Extend Canvas register drop handlers from the finder as urls using the UTI array to map candidate Node Classes. 
* Allow instantation of the Nodes via drag and drop.

tobyspark:
I’ve got this working, PR once I’ve tested it a bit more

vade:
closed via #210

PR #210, opened 2026-03-17
Feat – Drag onto document, implements #198

tobyspark:
> We want to allow for any nodes which support file loading via the File Loading node protocol extension to support auto instantiation of a new node with a populated file url path.
– #198 

This also makes the inspector UI fit for purpose. Aside – I like the idea of a grid-view multi-picker, which what was there was the remnants of, but we should get the basics right first. A separate node, I suspect, as we’d probably want to then get into pre-loading etc. for instantaneous switching, and that might take some more UI.

Note the first implementation of this extended `NodeFileLoadingProtocol` but this was over-wrought, requiring specialised init functions for these provider nodes when none were needed before. These are different to the nodes which _need_ files as part of their functionality, e.g. nodes loading shaders as intrinsic functionality. These nodes can, and were, instantiated normally and then the fileURL input can be set separately. So this PR implements a focussed `NodeFileDropTarget` protocol, which results in a good chunk less code. I could do the git archaeology to revive the original implementation if wanted.

### DIFF
--- a/Fabric/Views/Parameters/ParameterGroupView.swift
+++ b/Fabric/Views/Parameters/ParameterGroupView.swift
@@ -8,10 +8,12 @@
 
 import SwiftUI
 import Satin
+import UniformTypeIdentifiers
 
 struct ParameterGroupView : View
 {
     let parameterGroup:ParameterGroup
+    var fileContentTypes: [UTType] = [.data]
 
     var body: some View
     {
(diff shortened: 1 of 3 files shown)


tobyspark:
Gah, hit go a moment too soon. Have just insta-force-pushed with the updated message for the rebased commit.

tobyspark:
...for review.

Fabric/Graph/Node/Node.swift
```diff
@@ -601,7 +602,17 @@ import Combine
...
+/// Nodes that are constructed from a file (e.g. Metal shader effect nodes).
+public protocol NodeFileLoadingProtocol : Node
 {
     init(context:Context, fileURL:URL) throws
 }
+
+/// Nodes that accept a user-dropped file via a file-path parameter port.
+/// Conformers declare which UTTypes they handle and receive the URL after
+/// normal construction via ``setFileURL(_:)``.
+public protocol NodeFileDropTarget : Node
```
vade: id prefer extending `NodeFileLoadingProtocol` - can we factor out `NodeFileDropTarget`

Fabric/Nodes/NodeRegistry.swift
```diff
@@ -15,6 +16,32 @@ public class NodeRegistry {
...
+                }
+            }
+        }
+        return nil
+    }
+
+    /// All UTTypes accepted by drop-target nodes, for use with drop destination handlers.
+    public var allSupportedDropTypes: [UTType] {
+        var types: [UTType] = []
+        for nodeClass in self.nodesClasses {
```
vade: same as above.

Fabric/Views/Parameters/ParameterGroupView.swift
```diff
@@ -8,10 +8,12 @@
 
 import SwiftUI
 import Satin
+import UniformTypeIdentifiers
 
 struct ParameterGroupView : View
 {
     let parameterGroup:ParameterGroup
+    var fileContentTypes: [UTType] = [.data]
```
vade: Why do all instances of `ParameterGroupView` need a fileContentTypes var only set by `FileImportParameterView` ? Cant we move this to `FileImportParameterView` and be done with it?

vade:
Hey hey, 10k foot view, this is awesome and much needed improvements. I have just some minor gripes which i think make ParameterGroupView a bit cleaner and less leaky, otherwise this looks great, ill do a proper test later tonight!

vade:
Im fiddling on a branch to fix my nits - stay tuned. This is worth fast tracking :)

PR #223, opened 2026-03-28
Feat – Better new document

tobyspark:
Use the new Euler Orientation node and Auto-Layout feature to improve the default Fabric document.

### DIFF
--- a/Fabric Editor/FabricDocument.swift
+++ b/Fabric Editor/FabricDocument.swift
@@ -45,35 +45,58 @@ class FabricDocument: FileDocument
         print("Basic Document Init")
         let graph = Graph(context: self.context)
 
+        // Time source
+        let currentTimeNode = CurrentTimeNode(context: self.context)
+
+        // Math expression: secs * speed
+        let mathNode = MathExpressionNode(context: self.context, expression: "secs * speed")
+
+        // Publish the 'speed' port with a default of 10
+        let speedPort = mathNode.findPort(named: "speed", as: ParameterPort<Float>.self)!
+        speedPort.published = true
+        speedPort.value = 10
+
+        // Euler orientation (drives mesh rotation on X and Y)
+        let eulerNode = EulerOrientationNode(context: self.context)
+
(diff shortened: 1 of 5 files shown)


vade:
See comments in #224 about feelings on exposing to end users node settings derived logic at runtime.

tobyspark:
Yep, per #224 I’ll do a pass to add convenience initialisers for nodes with settings.

tobyspark:
This now has a commit to add convenience initialisers, and then a commit for the new document (force pushed)

vade (changes requested):

vade:
so in thinking about this a bit, i'd say lets only do the expression initializer for now so we can start with default expressions - 

I'm not convinced the current OSC and Timeline initializers are doing anything of interest at all and the Keyboard binding is in theory sensible but tbh  im just skeptical of this approach to custom node initializers in general... 

maybe ill get over it or see the light - as my biggest fear is this is going to snowball into a sidecar api that configures nodes at runtime not during initialization. the initializer should write to private properties that dont allow for public fiddling.

sorry to be a pain in the ass / stickler !

tobyspark:
The light is pretty bright – if you want to create graphs via API, initialisers to set the settings need to be part of that API. So the principle here is cut and dried to me.

Pragmatically, `Math Expression` we have hit already in Fabric Editor. `Switch` (aka multiplexer) I’ve hit in *spark stage (#191). `Sub Graph` I’ve hit in *spark stage, but that’s not settings per se, it’s setting the sub-graph (not in this PR, but perhaps should be).

I am not too worried about the specifics of these initialisers, as this is open-source and anyone programatically creating graphs will be capable of shaping the initialiser to what’s needed. But the pattern of convenience initialiser vs. making internal methods public needs to be clear. So my take is have the initialiser on any node with settings, but it could be a placeholder.

tobyspark:
Added a commit with fixes to the initialisers. Happy to also make the non-MathExp ones stubs, if it gets this in.

PR #229, opened 2026-03-31
Fix - `deltaTime` is computed correctly at source, workarounds removed

tobyspark:
The CAMetalDisplayLink callback computed deltaTime as previous - current, producing a negative value. Fixed to current - previous. Reverted the workaround in CAMetalDisplayLinkRenderer (commit e9c3be7) that computed its own delta, and removed the compensating negation in NumberIntegralNode.

Fixes #228


tobyspark:
A genie reviewed consumers of the timing and only found `NumberIntegralNode`.

PR #234, opened 2026-04-07
This PR adds improved settings panel for the LLM models, and removes some features from the node that are better configured in settings 

vade:
We add:

* updates MLX and MLX-Swift-LLM to latest versions, fixing #263
* a nicer UI with configuration settings and UI feedback for LLM Node and VLM Node including status, downloading info, etc
* improve memory management


vade:
Not quite ready yet, - WIP

vade:
This is in a good pretty spot from what i can tell, but there are some caveats

* model loading can take some time, even if model is already downloaded, depending on the size of the model
* model output is Async, and we dont *quite* have a nice way of dealing with that just yet.

PR #237, opened 2026-04-11
Feat – Parameter nodes for all

tobyspark:
Introduces `PassthroughNode<T>` and adopts it through the codebase.

In doing so
- addresses the UX inconsistency around “Insert Parameter Node” not being available for some ports.
- addresses the UX inconsistency between pure “splitter” type nodes and “make this type” nodes.
- adds special handling for color ports, which are not uniquely defined by their underlying value type.


Fabric/Graph/Port/PortType.swift
```diff
@@ -213,15 +213,38 @@ extension PortType {
...
-        case .Transform:    return IdentityTransformNode.self
+        case .Float:        return PassThroughNode<Float>.self
+        case .Int:          return PassThroughNode<Int>.self
+        case .Bool:         return PassThroughNode<Bool>.self
+        case .String:       return PassThroughNode<String>.self
+        case .Vector2:      return PassThroughNode<simd_float2>.self
+        case .Vector3:      return PassThroughNode<simd_float3>.self
+        case .Vector4:      return PassThroughNode<simd_float4>.self
+        case .Color:        return ColorPassThroughNode.self
+        case .Quaternion:   return PassThroughNode<simd_quatf>.self
```
vade, line 216: this shouldnt be a quatf, we pass quats though as simd float4s

Fabric/Nodes/Parameters/Color/MakeColorNode.swift
```diff
@@ -0,0 +1,56 @@
...
+    override public class var nodeTimeMode: Node.TimeMode { .None }
+    override public class var nodeDescription: String { "Converts 4 numerical components to a Color" }
+
+    // Ports
+    override public class func registerPorts(context: Context) -> [(name: String, port: Port)] {
+        let ports = super.registerPorts(context: context)
+
+        return ports +
+        [
+            ("inputR", ParameterPort(parameter: FloatParameter("R", 0.0, .inputfield, "Red component (0–1)"))),
```
vade, line 28: fabric is HDR compliant, so we support colors with HDR headroom values over 1.0 - not sure we need the range?

Fabric/Nodes/NodeRegistry.swift
```diff
@@ -110,6 +112,8 @@ public class NodeRegistry {
         
     private var materialNodeClasses:[Node.Type] = [
         // Materials
+        PassThroughNode<Material>.self,
+        PassThroughNode<Shader>.self,
```
vade: we dont need this, - sorry, shader is going to be removed as material kind of does what we need. shader should get removed as a viable port type / value representable or what not.

Fabric/Nodes/Parameters/Color/ColorPassThroughNode.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Created by Claude on 4/11/26.
+//
+
+import Foundation
+import Satin
+import simd
+import Metal
+
+/// Patching utility node for Color ports. Uses a color picker for the input.
+public class ColorPassThroughNode: Node
```
vade, line 14: can't we subclass PassthroughNode<simd_float4> and make this a pretty minimal super / override vs a larger node?

Fabric/Nodes/Parameters/Color/ColorPassThroughNode.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    // Port Proxy
+    public var input: NodePort<simd_float4> { port(named: "input") }
+    public var output: NodePort<simd_float4> { port(named: "output") }
+
+    public override func execute(context: GraphExecutionContext,
+                                 renderPassDescriptor: MTLRenderPassDescriptor,
+                                 commandBuffer: MTLCommandBuffer)
+    {
+        self.output.send(self.input.value)
```
vade: should check valueDidChange before sending

Fabric/Nodes/Parameters/PassThroughNode.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+/// Patching utility node that passes a value through without modification.
+/// Uses an editable parameter port for the input when the type supports it.
+public class PassThroughNode<T: PortValueRepresentable>: Node
+{
+    override public class var name: String { T.portType.rawValue }
+    override public class var nodeType: Node.NodeType { .Utility }
+    override public class var nodeExecutionMode: Node.ExecutionMode { .Provider }
```
vade, line 72: is there a reason this is a provider node? Adding input splitters as providers throughout the middle of the graph changes the execution dynamics. this should be a processor by default, or a flexible provider / processor depending on the input ports connection. 

no connection = provider, connection = processor?

vade (commented):
nice. 

some comments inline w questions 

one large one - i realize maybe PortValueRepresentable as a templated generic might be working against you - i wonder if this could be generic against the actual PORT TYPE (ie `public indirect enum PortType`  my thinking is it makes the semantics a bit easier and ameliorates switching - but i forget if that isnt doable due to swift bullshit. 

Just thinking aloud - not a major critique. 

Another note: I maybe regretting my decision to proxy orientation as simd float 4 as well as color (quaternion nodes should output simd float4's as the representation, as do color ... maybe making things a bit more slippery than needed?) 

curious your thoughts on either

* standardizing another type (quatf) as an output type? I dont know if we need another type making shit complicated and patching non obvious. 
* boosting color to its own type independent of simd_float4 (doing the same)

pro's it simplifies the graph port stuff

cons it makes for more port types that are incompatible when really it should 'just work'

Fabric/Nodes/Parameters/PassThroughNode.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    // Port Proxy
+    public var input: NodePort<T> { port(named: "input") }
+    public var output: NodePort<T> { port(named: "output") }
+
+    public override func execute(context: GraphExecutionContext,
+                                 renderPassDescriptor: MTLRenderPassDescriptor,
+                                 commandBuffer: MTLCommandBuffer)
+    {
+        self.output.send(self.input.value)
```
vade: i believe we should be checking valueDidChange as well no??

tobyspark:
> one large one - i realize maybe PortValueRepresentable as a templated generic might be working against you - i wonder if this could be generic against the actual PORT TYPE (ie public indirect enum PortType my thinking is it makes the semantics a bit easier and ameliorates switching - but i forget if that isnt doable due to swift bullshit.

Swift generics parameterise over types (<T: SomeProtocol>), not values. An enum case is a value, so it can't be a generic parameter.

> curious your thoughts on either

> standardizing another type (quatf) as an output type? I dont know if we need another type making shit complicated and patching non obvious.
> boosting color to its own type independent of simd_float4 (doing the same)

I think the UX of them being appearing different types is clearest (i.e. ports say color vs. vec4) and most obviously useful (i.e. you get a colour picker widget). What do we lose? The ability to take an output that is already vec4 but not colour/orientation as input for one? Is there an obvious case when that would happen?

vade:
Right, i vaguely recall running into the enum case != type generic barrier. Thanks for the clarification.

The color / port ui should already be differentiated  in the UI and description and should work since its based off of the PortType enum  - i was more concerned about 'hacks' in the code to differentiate different 'cases' for vec4 behavior, and weird side effects therein

tobyspark:
Another shared type with different UI: `boolean` as either a toggle or button.

tobyspark:
Pushed some naming clarifications. Having `<output type> From <input type>` rather than `Make <output type>` is more verbose in the simple cases, but soon becomes more compact and consistent (from value node aka input splitter up).

PR #238, opened 2026-04-11
Feat – Unified blend mode node

tobyspark:


Fabric/Nodes/Base/BaseImageNode.swift
```diff
@@ -463,16 +463,21 @@ public class BaseImageNode: Node, NodeFileLoadingProtocol
         return ports[index].value?.texture
     }
 
+    /// Parameter labels that material sync should not touch.
+    /// Subclasses can override this to protect ports they manage themselves.
+    open var materialSyncExcludedLabels: Set<String> { [] }
```
vade: Is this strictly needed? Do we actually have issues without it? This feels extremely fiddly to me.

Fabric/Nodes/Image/Mix/BlendNode.swift
```diff
@@ -0,0 +1,92 @@
...
+//
+//  Created by Claude on 4/11/26.
+//
+
+import Foundation
+import Satin
+import simd
+import Metal
+
+public class BlendNode: BaseImageNode
```
vade, line 13: Would highly suggest inverting the method here. Every single pixel hits a switch statement that isnt needed

I'd suggest a changing the loaded material / shader that is used in the pipeline dynamcially vs switching within the shaders processing code. 

this also ameliorates the need for changing BaseImageNode at all - i dont think theres a reason to touch it save for using this method

Fabric/Shaders/BlendShader.metal
```diff
@@ -0,0 +1,75 @@
...
+
+typedef struct {
+    float mode;   // input, 0, Mode
+    float amount; // slider, 0.0, 1.0, 0.0, Amount
+} BlendUniforms;
+
+half3 applyBlendMode(int mode, half3 a, half3 b)
+{
+    switch (mode)
+    {
```
vade: Not a fan of switch statements in shaders :(

Fabric/Nodes/Image/Mix/BlendNode.swift
vade: Also note that Satin has a shader cache, so we shouldnt worry about recompilation / loading.

vade (commented):
Do you mind re-working this with the genie slightly?

Fabric/Nodes/Image/Mix/BlendNode.swift
tobyspark: Agreed.

Fabric/Nodes/Base/BaseImageNode.swift
tobyspark: Yes it’s needed. Without it serialisation breaks. There were already hardcoded overrides for `width` and `height`. This makes that available on a node by node basis.

vade: On main we dont have `materialSyncExcludedLabels` and serialization works via port serialization - thats not needed - BaseImage node serializes just fine.

Fabric/Nodes/Image/Mix/BlendNode.swift
tobyspark: Resolved in 4f97532ac1d241fd2ad9f9e91986b3b1816dd8af

Fabric/Nodes/Base/BaseImageNode.swift
vade: I think this was about avoiding 'name collisions' with ports, but thats also not strictly an issue for us, port serialization doesnt have that issue.

tobyspark: In `main`, `syncDynamicParameterPortsFromMaterial` removes any parameter port without a derived-from-material label. That will remove the `mode` port. You can add `mode` dynamically, but then you get into serialisation issues as the value is set on deserialisation but then ripped out by the sync. Much better to genericise the hardcoded carve-out for `width` and `height` that actually was in `syncDynamicParameterPortsFromMaterial` all along.

vade: * when are ports changing in Blend Mode? 
* port is registered via the registration system, which handles serialization
* width and height deserialize with no issue in nodes which are generators for example. 

can you show me a concrete failure path in existing serialization or nodes?

secondly if `syncDynamicParameterPortsFromMaterial` has a bug, my suggestion would be to not add a new method to the class, but to fix the bug in `syncDynamicParameterPortsFromMaterial`

vade: Also, general best practice with PRs and a request, can you summarize the changes made in the PR, and if any root classes / existing classes have changes, please note if its a bug, new behavior required or a refactor etc. 

I have to do a ton of work to manually review the code, and no PR description means im trying to reason about things and understand whats what and why.

tobyspark:
> Also, general best practice with PRs and a request, can you summarize the changes made in the PR, and if any root classes / existing classes have changes, please note if its a bug, new behavior required or a refactor etc.
>
> I have to do a ton of work to manually review the code, and no PR description means im trying to reason about things and understand whats what and why.

For PRs to not get onerous for both parties, I strongly think we should commit to the weekly review idea. It’s going to be so much more efficient in conversation, with shared context up in front of us both.

tobyspark:
From call – can we remove `materialSyncExcludedLabels` entirely (and the width/height before it) by using some set arithmatic. I will do...

Fabric/Nodes/Base/BaseImageNode.swift
tobyspark: Resolved via call and today’s commits

tobyspark:
@vade Ready for a final look / merge.

vade:
Thanks for your patience!

PR #239, opened 2026-04-15
Add MSDF text image and world text billboard nodes

jvcleave:
Proof of concept for MSDF generated bitmaps for metal rendering

tool for generating files is here

https://github.com/jvcleave/MSDFTextEngineTool/

TextEngineToolApp is the easiest route but you still have to build the TextEngineTool. see the readmes there

I have attached HelveticaNeue

2 node examples, TextImageNode, WorldTextBillboardNode

add them to the graph and point the inspector at the atlas-runtime json and bmp

screenie example
image

I've also added a TODO for TextEngineToolApp to generate Satin compatible atlas files so might be interesting to compare at some point (or update Satin to use 3 channels)

[HelveticaNeue.zip](https://github.com/user-attachments/files/26741143/HelveticaNeue.zip)

### DIFF
--- a/Fabric/Nodes/Image/TextImageNode.swift
+++ b/Fabric/Nodes/Image/TextImageNode.swift
@@ -1,0 +1,542 @@
+import Foundation
+import Satin
+import simd
+import Metal
+import MetalKit
+import ImageIO
+
+public final class TextImageNode: Node
+{
+    public override class var name: String { "Text Image" }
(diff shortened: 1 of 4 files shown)


vade:
Hey this is awesome - thank you! I feel like having 3D Text SDF is better placed in Satin mainline first, and then can be exposed as a Satin Material similar to the single channel text SDF path it has currently?

Issue #61, opened 2025-10-29
Migrate Node Addition to proxy through GraphRenderer


vade:
To keep separate of responsibilities intact, due to the fact nodes are now implementing `start/stop/enable/disable execution` requires us to have a API contract that can provide a `GraphContext` appropriately

Today our Graph can simply provide a `Satin.Context`

So we either

1) refactor `start/stop/enable/disable execution` to only consume a `Satin.Context` which might have some side effects
2) move node additions and removals to go through a mediator - the `GraphRenderer` which can provide said `GraphContext`

Programming is boring.

vade:
Fixed in Fabric / Satin 2.0 branch

PR #245, opened 2026-05-10
AudioSpectrumNode Fixes & Improvements

tobyspark:
Inspector / behaviour:

- Replace the original `Gain` input with `Sensitivity` (drives the noise floor / how responsive the analyser is to quiet sounds) plus an overdrive `Gain` that multiplies normalised bar values for visual punch without touching the underlying signal. Adds a `BandsVisualizer` settings popover so authors can see what the analyser is producing inline.
- Wire the `Device Name` dropdown to actually route the capture session to the selected `AVCaptureDevice`, instead of always using the system default input.
- Rename the output port display name from "Number" to "Spectrum" so it reads correctly in connection menus.

Threading + state:

- Wrap the device list in `Mutex<DeviceList>`. The capture-session setup pulls `devices` from the embedder's consumer queue while `AVCaptureDevice` connect/disconnect notifications mutate it from main; the lock makes the cross-queue access safe.
- Move `samples` access behind a `Mutex` and drop the `DispatchQueue.main.async` hop in the capture callback. The capture thread now only stages raw samples + the latest sample rate, and `execute()` is the sole site that drains them and mutates filter-bank state. Works regardless of which queue the embedder runs the graph on.

Fixes:

- Engine wasn't starting for newly-added nodes. Addressed by deferring filter-bank creation until the first real audio buffer arrives.
- NaN poisoning of the biquad filter bank: `AVAudioEngine.inputNode. inputFormat` returns `sampleRate=0` on macOS before the engine is running, which poisoned every biquad coefficient with NaN and never recovered. Build/rebuild the filter bank from the actual stream rate observed in the tap callback, gated on `bufferSampleRate.isFinite && > 0`. Self-heal on the envelope side too: if `env[k]` becomes non-finite from a transient, reset it to 0 instead of letting NaN propagate forever.

### DIFF
--- a/Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
+++ b/Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
@@ -235,7 +235,7 @@ public class AudioSpectrumNode : Node
     // selection via AVCaptureDeviceInput is the supported macOS API for
     // per-node device switching, where AVAudioEngine's AUHAL-routing is not.
     @ObservationIgnored private var captureSession = AVCaptureSession()
-    @ObservationIgnored private let captureQueue = DispatchQueue(label: "net.sparklive.AudioSpectrum.capture")
+    @ObservationIgnored private let captureQueue = DispatchQueue(label: "fabric.AudioSpectrumNode.capture_queue")
     @ObservationIgnored private var captureDelegate = CaptureDelegate()
 
     // GraphRenderer only calls startExecution once, at renderer setup. Nodes
--- a/Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
+++ b/Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
@@ -12,6 +12,7 @@ import simd
 import AVFoundation
 import Accelerate
 import Dispatch
+import Synchronization
 
 public class AudioSpectrumNode : Node
 {
@@ -237,13 +238,12 @@ public class AudioSpectrumNode : Node
     @ObservationIgnored private let captureQueue = DispatchQueue(label: "fabric.AudioSpectrumNode.capture_queue")
     @ObservationIgnored private var captureDelegate = CaptureDelegate()
 
-    /// Capture-thread → consumer-thread handoff. The capture callback runs
-    /// on `captureQueue` and appends raw samples + the latest observed
-    /// sample rate to these properties directly. `execute()` runs on the
-    /// embedder's queue (main in Fabric Editor, a private serial queue in
-    /// Spark Stage) and drains via `captureQueue.sync { … }`. All
-    /// node-state mutation (filter-bank rebuild, etc.) happens on the
-    /// consumer thread.
+    /// Capture-thread → consumer-thread handoff. The capture callback runs on
+    /// `captureQueue` and appends raw samples + the latest observed sample
+    /// rate to these properties directly. `execute()` runs on the embedder's
+    /// queue (main in Fabric Editor, other embedders may use e.g. a private
+    /// serial queue) and drains via `captureQueue.sync { … }`. All node-state
+    /// mutation (filter-bank rebuild, etc.) happens on the consumer thread.
     @ObservationIgnored private var pendingSamples: [Float] = []
     @ObservationIgnored private var lastSeenSampleRate: Float?
 
@@ -255,10 +255,34 @@ public class AudioSpectrumNode : Node
         mediaType: .audio,
         position: .unspecified
     )
-    /// Available audio devices. Written from the main queue (via
-    /// `AVCaptureDevice` connect/disconnect notifications) and read from
-    /// the embedder's consumer queue (via `setupCaptureSession` →
-    /// `resolveSelectedAudioDevice`). Mutex provides cross-queue safety.
+    
+    /// Available audio devices, populated from `AVCaptureDevice`
+    /// connect/disconnect notifications.
+    ///
+    /// The observer is registered with `queue: .main` because the
+    /// handler updates the dropdown's `StringParameter.options`, and
+    /// Fabric's `@Observable` engine types (Node, Parameter, …) are
+    /// main-thread-affine: Observation's change tracking misbehaves on
+    /// non-main mutations. That's a property of Fabric's current
+    /// parameter design — engine state and UI binding state share a
+    /// single object — not a SwiftUI requirement of this node. A
+    /// UI-agnostic factoring would lift the observable wrapper out of
+    /// the engine layer and let nodes own their state on the consumer
+    /// queue, removing the `.main` constraint and the cross-queue case
+    /// it creates.
+    ///
+    /// Given that constraint: writes happen on main, reads happen on
+    /// the embedder's consumer queue (main in the Fabric Editor, a
+    /// private serial queue elsewhere). When the consumer queue is
+    /// main, both collapse to one thread and a lock is redundant —
+    /// that's the path `CameraProviderNode` and
+    /// `ScreenCaptureProviderNode` take, with a plain
+    /// `var devices: [AVCaptureDevice]`. Under a non-main consumer
+    /// queue it's a genuine cross-queue read; the race is benign for
+    /// short reference-array stores (stale dropdown matching, not
+    /// crashes), but this node uses a Mutex to make the hand-off
+    /// explicit rather than rely on the project's tolerated-race
+    /// convention.
     private struct DeviceList: ~Copyable, @unchecked Sendable {
         var devices: [AVCaptureDevice] = []
     }
--- a/Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
+++ b/Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
@@ -12,7 +12,6 @@ import simd
 import AVFoundation
 import Accelerate
 import Dispatch
-import Synchronization
 
 public class AudioSpectrumNode : Node
 {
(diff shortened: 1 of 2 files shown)


Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
+        {
+            self.owner?.processAudioSampleBuffer(sampleBuffer)
+        }
+    }
+
+    // AVCaptureSession replaces AVAudioEngine: its per-session input device
+    // selection via AVCaptureDeviceInput is the supported macOS API for
+    // per-node device switching, where AVAudioEngine's AUHAL-routing is not.
+    @ObservationIgnored private var captureSession = AVCaptureSession()
+    @ObservationIgnored private let captureQueue = DispatchQueue(label: "net.sparklive.AudioSpectrum.capture")
```
vade: `graphics.fabric.AudioSpectrum.capturequeue` :) hehe

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
+    @ObservationIgnored private var captureDelegate = CaptureDelegate()
+
+    // GraphRenderer only calls startExecution once, at renderer setup. Nodes
```
vade: This wont be true, and we should be calling startExecution and enableExecution on new node adds to a graph. Lets not work around these issues per node, if renderer has a bug we should fix it and not have too much side car behaviour sprinkled around.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
+        position: .unspecified
+    )
+    /// Available audio devices. Written from the main queue (via
+    /// `AVCaptureDevice` connect/disconnect notifications) and read from
+    /// the embedder's consumer queue (via `setupCaptureSession` →
+    /// `resolveSelectedAudioDevice`). Mutex provides cross-queue safety.
+    private struct DeviceList: ~Copyable, @unchecked Sendable {
+        var devices: [AVCaptureDevice] = []
+    }
+    @ObservationIgnored private let devicesLock = Mutex<DeviceList>(DeviceList())
```
vade, line 289: if we have queues we do not need a lock - just dispatch sync to the owning queue.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
+            forName: AVCaptureDevice.wasDisconnectedNotification, object: nil, queue: .main
+        ) { [weak self] _ in
+            self?.refreshAudioDeviceOptions()
+        }
+    }
+
+    private func refreshAudioDeviceOptions()
+    {
+        let fresh = self.discoverySession.devices
+        self.devicesLock.withLock { $0.devices = fresh }
```
vade, line 331: could be dispatch sync

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
-            case .notDetermined: // The user has not yet been asked for camera access.
-                AVCaptureDevice.requestAccess(for: .audio) { granted in
+            case .authorized:
+                self.setupCaptureSession()
+            case .notDetermined:
+                AVCaptureDevice.requestAccess(for: .audio) { [weak self] granted in
                     if granted {
                         print("Granted Mic Access")
-                        self.setupAudioShit()
+                        DispatchQueue.main.async { self?.setupCaptureSession() }
```
vade: ah nice catch!

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
+        self.didRequestAudioSetup = true
 
         switch AVCaptureDevice.authorizationStatus(for: .audio) {
-            case .authorized: // The user has previously granted access to the camera.
-                self.setupAudioShit()
-                return
-            case .notDetermined: // The user has not yet been asked for camera access.
-                AVCaptureDevice.requestAccess(for: .audio) { granted in
+            case .authorized:
+                self.setupCaptureSession()
```
vade, line 357: shouldnt this also have the main queue async dispatch from below?

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,91 +188,215 @@ public class AudioSpectrumNode : Node
...
+    @ObservationIgnored private var didRequestAudioSetup = false
+
+    /// Capture-thread → consumer-thread handoff. The capture callback runs on
+    /// `captureQueue`; `execute()` runs on whatever queue the embedder uses
+    /// (main in Fabric Editor, a private serial queue in Spark Stage). The
+    /// capture thread *only* stages raw samples + the latest observed sample
+    /// rate here; all node-state mutation (filter-bank rebuild, etc.) happens
+    /// on the consumer thread inside `execute()`. This means the producer is
+    /// embedder-agnostic — no need to know which queue runs Fabric.
+    private struct AudioBuffer: ~Copyable {
```
vade: why do we need this to be copyable? Cant we use a functional paradigm and either have a pool or lock free hand off of data ?

vade (changes requested):
Thanks for tackling this! Theres a lot in here. This is a first cursory glance so take this review with a grain of salt!

My biggest question is - if we have queues, why do we need locks? Queues can act as locks by enforcing serial access to data from a specific context and removing data races, and I'm not convinced high level the locks are needed / required if we have the queues. 

Maybe at audio data rate they are faster than GCD dispatch? LMK your thoughts. 

Secondly, even with diff threads, can't we do lock free  by using producer / consumer model or maybe a more functional design of allocate and pass the data through through the filter bank ? 

Thanks!

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: ha! yep.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: I agree with the approach of fixing cause vs. symptom. 

Flagging that blocks this PR until the core implemenation is done, #61

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: I will do a consistency pass and test – working in both Fabric’s main and Spark Stage’s serial queue.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: I will do a consistency pass and test – working in both Fabric’s main and Spark Stage’s serial queue.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: I will do a consistency pass and test – working in both Fabric’s main and Spark Stage’s serial queue.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: Note it’s marked as _not_ Copyable. That prevents anyone from accessing the protected state without holding the lock. For the rationale on locks etc., I’ll reply as a main comment.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
tobyspark: For the rationale on locks etc., I’ll reply as a main comment.

vade: I did a bit of research, and apparently using GCD queues as locks was never really meant for fine grained locking, but for broader tasks. theres an order of magnitude or two ( !) or more latency in getting a mutex vs the GCD infra - so thats worth knowing! :D

tobyspark:
On queues and locks –

AVCaptureAudioDataOutput works with a queue, so typical RT concerns of no locks etc. don’t apply. So why then have a lock? In combination with the suppression of the default Copyable conformance, it’s a way to ensure thread safety at compile time. And it’s faster. 

We could go lock free with a circular buffer, but that’s more code complexity in a different direction, without the compile time safety.

I don’t think either the extra Swift complexity or possible performance gain is worth breaking the norms of Fabric elsewhere. Given a queue is already in there, let’s just `.sync` it.

With a similar rationale but for a much simpler case, scaling back from `Mutex<DeviceList>` introduces a cross-queue read in any embedder running Fabric off main thread, but its safe in practice here.

(Though if you want to push the Fabric idiom to be more defensive with type enforced locking etc., digging into this might be interesting as a starting point. And standard disclaimer: I’m no expert here)

vade:
is that an llm response? :D

vade:
Via Discord, now that the fix for #61 is in Feature/Satin 2.0 support, if you could target that branch for this PR, along with the minor changes to execute / enable / disable / start / stop exection function signatures, that would be awesome!

tobyspark:
These commits should resolve all comments except for Mutex on the DeviceList, for which the rationale is in the code comment (and points at a Fabric design issue that’s out of scope here).

I’ve changed the base branch for this PR to the Satin 2.0 branch, and now there are conflicts. I’ll get to those... in a bit.

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,94 +188,224 @@ public class AudioSpectrumNode : Node
...
+    @ObservationIgnored private let discoverySession = AVCaptureDevice.DiscoverySession(
+        deviceTypes: [.microphone, .external],
+        mediaType: .audio,
+        position: .unspecified
+    )
     
+    /// Available audio devices, populated from `AVCaptureDevice`
+    /// connect/disconnect notifications.
+    ///
+    /// The observer is registered with `queue: .main` because the
```
vade, line 262: I think this LLM slop documentation is wrong and misleading as far as I can tell?

The observer could absolutely use its own queue, and dispatch to the node to set on main.  These sorts of comments littered in the code base will start to skew other downstream LLM agents reads of the code and compound to bad decisions and a sort of 'look ma no hands' scenario and the accompanying context drift + reasoning collapse. 

Can we cut this down into a succinct, no story telling, just the fact's on the threading model without editorialization?

Fabric/Nodes/Parameters/Number/AudioSpectrumNode.swift
```diff
@@ -184,94 +188,224 @@ public class AudioSpectrumNode : Node
...
         switch AVCaptureDevice.authorizationStatus(for: .audio) {
-            case .authorized: // The user has previously granted access to the camera.
-                self.setupAudioShit()
-                return
-            case .notDetermined: // The user has not yet been asked for camera access.
-                AVCaptureDevice.requestAccess(for: .audio) { granted in
+            case .authorized:
+                self.setupCaptureSession()
+            case .notDetermined:
+                AVCaptureDevice.requestAccess(for: .audio) { [weak self] granted in
```
vade, line 359: `        The completion handler is called on an arbitrary dispatch queue. It is the client's responsibility to ensure that any UIKit-related updates are called on the main queue or main thread as a result.`

this should be punted to main

vade (commented):
One threading issue, one probably annoying comment by me on LLM comments - thanks for bearing with me.

vade:
Gonna merge th

PR #246, opened 2026-05-10
Movie Provider Improvements

tobyspark:
- Adds `Playing`, `Seek Time`, and `Seek Behaviour` inputs to the Movie Provider node
- Adds basic HAP codec support to the Movie Provider node

Notes
- I’m not set on `Seek Behaviour`. If all videos were HAP or similar all-keyframe codecs, this would not be needed. But given Fabric can be used in a realtime context, any use of H264 etc. does I think require a decision between instant response and accurate response. Do we expose the actual underlying API, which is a time interval? Or do what I tried here which is expose our intent, at the cost of mis-characterising what’s happening (and may not even be true if there’s some other codec / playback issue I haven’t thought of).
- This does not (yet) support all HAP codecs, e.g. ones that require post-processing. That’s more work to implement and test. Do we want to do that here?
- This is gated to macOS and mirrors the Syphon framework installation. But perhaps Syphon nodes and a Movie provider with HAP support is something that should be pushed out to plugins (when we get there).


vade:
So I think a design goal that i havent adequately documented is an Async prototol for nodes which can do operations async in a realtime context, but switch a single flag and do things synchronously for a 'offline rendering' context. 

My feeling is 

* file loading
* frame decode
* seek
* play pause

should be hidden behind that switch/  protocol support (meaning diff sync vs async behavior for depending on current  mode), but we dont need to solve that just yet I dont think. 

edit: added issue for that

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -53,27 +56,102 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
     override public class var nodeTimeMode: Node.TimeMode { .TimeBase }
     override public class var nodeDescription: String { "Play a Movie File from disk, providing a stream of output Images"}
 
+    // Seek behaviour options. `frame` is exact (zero-tolerance) seek;
```
vade: nit pick - things like this should be enums in my experience, it allows for clean switch statements  and doesnt need to rely on string matching.

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -53,27 +56,102 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+    /// changes in `execute`. Tolerance is selected from
+    /// `inputSeekBehaviourParam`. Re-primes playback (`player.play()`)
+    /// when the user wants the player playing — some seeks (notably
+    /// zero-tolerance ones) leave `rate` at 0 momentarily, which would
+    /// otherwise stall the player at the seek target.
+    private func performSeek(to seconds: TimeInterval)
+    {
+        guard self.player.currentItem != nil else { return }
+        let clamped = max(0, min(seconds, self.duration))
+        let target = CMTime(seconds: clamped, preferredTimescale: 600)
```
vade: nit pick from experience - this should probably use the natural time scale of the video track to get better frame accurate seeks

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -115,15 +193,78 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+            }
+        }
+
+        // Honour seek port. Negative values are the sentinel default —
+        // ignored so the asset isn't seeked to 0 on first load. Skip
+        // when the player is already at the target (a same-target write
+        // arriving while the player happens to be there is a no-op),
+        // but always seek when the player's actual position differs.
+        // This is what makes loop-back enforcement work: every time the
+        // player crosses `outPoint`, the embedder re-writes `inPoint` to
```
vade: i dont understand the context of the 'embedder' here - is this spark stage specific logic here? is this meant to debounce too many seeks that are close to one another? where does 0.05 come from?

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -133,8 +274,189 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
                 self.outputTexturePort.send( image )
             }
         }
+        else if self.player.rate == 0,
```
vade: Help me understand the logic here - why do we need to supply new frames for the same frame if rate is zero - the output port should hold the last image. Im probably missing some nuance here, but im curious why we need this case handled so specifically.

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -115,15 +193,78 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+            let target = TimeInterval(raw)
+            if abs(target - self.currentTime) > 0.05
+            {
+                performSeek(to: target)
+            }
+        }
+
+        let time = context.timing.time
+
+#if FABRIC_HAP_ENABLED
```
vade, line 355: Nit pick - can we extract the hap pixel buffer handling into its own functions and a single if / def and not split it into sections within the execute function?

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -133,8 +274,189 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+            if let pixelBuffer = self.playerItemVideoOutput.copyPixelBuffer(forItemTime: pausedTime, itemTimeForDisplay: nil),
+               let renderer = context.graphRenderer,
+               let image = renderer.newImage(fromPixelBuffer: pixelBuffer)
+            {
+                self.outputTexturePort.send( image )
+            }
+        }
      }
 
+#if FABRIC_HAP_ENABLED
```
vade: Same as above : Nit pick - can we extract the hap pixel buffer handling into its own functions and a single if / def and not split it into sections within the execute function?

vade: Ah, i read this wrong in the git diff, ignore!

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -53,27 +56,102 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+
+    /// `true` while a seek is in flight. Read-only. Embedders should
+    /// avoid writing fresh values to `inputSeekTimeParam` while this
+    /// is set — a player mid-seek still reports a stale `currentTime`
+    /// briefly, and re-driving the seek port on every frame would
+    /// cancel the previous seek and stall playback at the boundary.
+    public var isSeeking: Bool { self.seeking }
+
+    @ObservationIgnored private var seeking: Bool = false
+
```
vade, line 180: why not `public private(set) var isSeeking:Bool = false` and remove the proxied @ObservationIgnored private var?

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -115,15 +193,78 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+        // but always seek when the player's actual position differs.
+        // This is what makes loop-back enforcement work: every time the
+        // player crosses `outPoint`, the embedder re-writes `inPoint` to
+        // this port and we genuinely re-seek.
+        if self.inputSeekTimeParam.valueDidChange,
+           let raw = self.inputSeekTimeParam.value,
+           raw >= 0
+        {
+            let target = TimeInterval(raw)
+            if abs(target - self.currentTime) > 0.05
```
vade: does it make sense to gate against 'isSeeking' here? ie if we already have a seek in progress we havent finished, why seek again? Not sure it makes sense, just riffing.

vade (changes requested):
Overall this looks great, mostly q's and nit picks. Thank you!

Fabric/Nodes/Image/MovieProviderNode.swift
vade: read this wrong - git diff context confused me, ignore.

vade:
When you can, also review this branch to point to Feature/Satin 2.0 PR and update the node signatures similar to the audio one if you can

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: Agreed on enums. The feature here I have changed my mind on, see later.

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: Yeah this is historical cruft that shouldn’t be in here.

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: Ah, we shouldn’t. It should emit the seeked-to frame if paused, once, but that’s it.

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: Can do

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: There’s a threading reason why it’s like that, but actually this now moot because I hadn’t ensured the improvements coalesced seeks, which makes `isSeeking` purely an internal detail.

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: As alluded to, the seek handling wasn’t what it should have been, which is silly as I’ve implemented the correct pattern before.

tobyspark:
Minor things all addressed. Two bigger things of note
- I wasn’t happy with the ‘Seek Behaviour’ input / behaviour. Not controversial there’s a use-case for both. But pending the work on async vs. sync execution I think we only need to expose it as a custom initialiser as it’s something the host app will have the context to set the correct value for. And the default of `tolerance: .infinity` works for realtime use for now, as if you want keyframe behaviour use HAP not H264 or friends. (Perhaps this means it should be a node setting – happy to do that but easier to defer a final design until sync. vs async).
- The correct pattern for seeking is to seek to the first new time given, and wait for that seek to finish before seeking to any further time that arrived while seeking (I can’t remember the source, but it was authoritative at the time, and unlikely AVFoundation has changed behaviour since). Given I’ve implemented that by hand before, it’s a bit embarrassing not to have done that here despite the original work here being expedient in a push for a gig.

Now merging in Satin 2.0 branch...

tobyspark:
Merged Satin 2.0 in.

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -53,58 +56,201 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+        // is available — e.g. asset still loading.
+        let timescale = self.asset?.tracks(withMediaType: .video).first?.naturalTimeScale ?? 600
+        let target = CMTime(seconds: clamped, preferredTimescale: timescale)
+        let tol: CMTime = self.seekTolerance.isInfinite
+            ? .positiveInfinity
+            : CMTime(seconds: self.seekTolerance, preferredTimescale: timescale)
+        self.seeking = true
+        self.player.seek(to: target, toleranceBefore: tol, toleranceAfter: tol) { [weak self] _ in
+            guard let self else { return }
+            self.seeking = false
```
vade, line 232: strictly speaking i think this is a race condition. You dont control the queue that seek calls the closure on, so i believe theres a chance is isnt main, which means self's properties could be mutated out from under you, ie self.seeking, self.player.rate should be accessed on main imo, even if its maybe in practice  v rarely an issue?

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -53,58 +56,201 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+            // If the player is paused at seek-land time,
+            // AVPlayerItemVideoOutput won't emit a fresh pixel buffer
+            // on its own — flag the next execute() to push the seeked
+            // frame downstream as a one-shot.
+            if self.player.rate == 0
+            {
+                self.needsEmitAfterSeek = true
+            }
+        }
+        if (self.inputPlayingParam.value ?? true)
```
vade, line 242: why does perform seek need a play check?

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -133,8 +317,229 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+            let pausedTime = item.currentTime()
+            if let pixelBuffer = self.playerItemVideoOutput.copyPixelBuffer(forItemTime: pausedTime, itemTimeForDisplay: nil),
+               let image = renderer.newImage(fromPixelBuffer: pixelBuffer)
+            {
+                self.outputTexturePort.send( image )
+            }
+        }
      }
 
+#if FABRIC_HAP_ENABLED
```
vade: I havent looked at HAP implementations in a long time - im gonna trust this is referenced / pulled from a canonical example, and referenced, and not halucinated by Claude from its reading? How vetted is this?

Fabric/Nodes/Image/MovieProviderNode.swift
```diff
@@ -133,8 +317,229 @@ public class MovieProviderNode : Node, NodeFileLoadingProtocol
...
+
+    /// Performant path: turn a HapDecoderFrame's pre-decompressed DXT
+    /// bytes into a compressed MTLTexture in one upload. No RGB
+    /// expansion, no memcpy of decompressed pixels.
+    ///
+    /// Trade-off: a fresh MTLTexture is allocated per frame. Metal's
+    /// allocator handles this well at 60fps for typical Hap frame
+    /// sizes (~0.5–2MB compressed); a texture pool would be a
+    /// follow-up optimisation if profiling shows allocator pressure.
+    private static func makeDXTImage(fromHapFrame frame: HapDecoderFrame, device: MTLDevice) -> FabricImage? {
```
vade: A nice to have - Fabric's Graph Renderer has a new Image function which uses a metal memory heap. It might be nice to use the already allocated heap memory to back the HAP frame. Maybe a note we make an issue for in the future? Wanted to point it out.

vade (commented):
Looks mostly good to me. One question, one minor threading concern, and one nice to have. LMK!

Fabric/Nodes/Image/MovieProviderNode.swift
vade: yea i think you can use `renderer.sharedTextureCache.newManagedImage(...) and use the pixel format and descriptotor you have, and replace the device.makeImage call and use this and get better perf basically for free.

pass in the GraphRenderer to : 

```
    private static func makeDXTImage(fromHapFrame frame: HapDecoderFrame, device: MTLDevice) -> FabricImage? {
```

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: Yes it is a race condition, and yes in my use has not been an issue (and in theory the data type would be atomic on Arm*). 

What we shouldn't do is assume or require Fabric to run on main, if we’re serious about it being an embeddable rendering engine (e.g. *spark stage runs Fabric on a private serial queue, e.g. [related issue](https://github.com/Fabric-Project/Fabric/pull/245/changes/1212bb238630f991285d5b16da134738a86451b0) in Audio PR).

I think there are two options
- Introduce a lock
- Gain a reference to the queue that `execute` was called on, and pass the work there.

Lock is a ~~one~~few-liner. The queue pass-off can probably be hacked, but feels like there’s some architectural agreement on Fabric itself that should happen before even thinking about that.

Whaddyreckon?

\* Standard caveats apply.

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: This is explained in the function comment. The check is to not un-pause paused playback.

The comment could be updated to something like
```
// Defensive: a seek can leave the player either with rate=0 (Apple
// acknowledges this for scrub/seek/scan operations, forums thread 663489)
// or in .waitingToPlayAtSpecifiedRate. Calling play() re-asserts the
// requested rate and re-primes buffering.
```

Fabric/Nodes/Image/MovieProviderNode.swift
tobyspark: Nice. \<goes does\>

tobyspark: implemented in [219436](https://github.com/Fabric-Project/Fabric/pull/246/commits/2194936c3574993eb81c242b6d2e3d7e21003f36)

tobyspark:
Pushed the texture pool enhancement and further clean-ups

Fabric/Nodes/Image/MovieProviderNode.swift
vade: I dont see that explained anywhere about that specific behavior and it is factually wrong hallucinated BS, and the thread absolutely does not say that a scrub can leave the player in a `. waitingToPlayAtSpecifiedRate` - in fact, `waitingToPlayAtSpecifiedRate` isnt mentioned once in that thread - did you read it? 

the only thing that thread implies is that during a fast forward event rate may go to zero as seek is used, which pauses internally. 

i worry we are trusting these stupid fucking LLMs a bit too much and taking their input at face value. Working with them in my experience requires a lot of critical 'what in the ever living fuck are you talking' about language and push back on added complexities. Remember they are trained on the worlds code, most of which is total fucking garbage. 

I know they are empowering and its enticing, but absolutely do not trust them, and review code manually for bullshit, and as them to critique, proof check claims, simplify and iterate on the outcomes a bit. This stuff is really hard to do ex post facto.

vade: One technique ive found that is particularly helpful is finding canonical reference implementations, it helps them not go off the rails and come up with poorly reasonsed solutions that mix different ideas together.

Fabric/Nodes/Image/MovieProviderNode.swift
vade: and btw, my more recent comments are kind of why i was asking about checking this

Here are claudes critiques of claudes own code, just for reference, and even that has some incoherent stuff in it

```
---                                                                                                                                                                                                                                                                  
  1. Asset tracks race condition (significant)                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                       
  In loadAssetFromInputValue, containsHapVideoTrack() is called immediately after creating the AVURLAsset:                                                                                                                                                             
                                                                                                                                                                                                                                                                       
  self.asset = AVURLAsset(url: url, options: [...])                                                                                                                                                                                                                    
  // ...                                                                                                                                                                                                                                                               
  if self.asset?.containsHapVideoTrack() == true,                     
                                                                                                                                                                                                                                                                       
  AVURLAsset loads tracks asynchronously. At this point asset.tracks is almost certainly empty — containsHapVideoTrack() will return false for any freshly-created asset including valid Hap files. The automaticallyLoadedAssetKeys on the AVPlayerItem triggers async
   loading but doesn't block here. The HAP path will never be entered for files loaded cold.                                                                                                                                                                           
                                                                                                                                                                                                                                                                       
  ---                                                                                                                                                                                                                                                                  
  2. allocFrameClosest(to:) called unconditionally every frame                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                       
  guard let frame = hapOutput.allocFrameClosest(to: itemTime) else { return true }
                                                                                                                                                                                                                                                                       
  The standard path checks hasNewPixelBuffer(forItemTime:) before copying. The HAP path skips any equivalent check and allocates a frame object on every execute() call — 60 times per second whether or not the video has advanced. If AVPlayerItemHapDXTOutput       
  exposes a hasNewFrame(forItemTime:) or equivalent, it should gate this allocation.                                                                                                                                                                                   
                                                                                                                                                                                                                                                                       
  ---                                                                 
  3. Per-frame MTLTexture allocation in the DXT fast path
                                                                                                                                                                                                                                                                       
  The comment acknowledges this but dismisses it. device.makeTexture(descriptor:) at 60fps for a 1080p Hap1 frame means 60 allocations/sec of ~0.5–2MB GPU-mapped memory. Metal's allocator handles this, but it's the main cost you're trying to avoid by using the
  DXT path at all. A two- or three-slot texture pool (index by frame modulo) would eliminate this entirely and is not complex.                                                                                                                                         
                                                                      
  ---                                                                                                                                                                                                                                                                  
  4. Removing the wrong output from old items                         
                                                                                                                                                                                                                                                                       
  if let oldItem = self.player.currentItem
  {                                                                                                                                                                                                                                                                    
      oldItem.remove(self.playerItemVideoOutput)  // always called... 
      if let oldHap = self.hapOutput {                                                                                                                                                                                                                                 
          oldItem.remove(oldHap)                 // ...even if this was actually attached
      }                                                                                                                                                                                                                                                                
  }                                                                   
                                                                                                                                                                                                                                                                       
  When the previously loaded asset was Hap, playerItemVideoOutput was never added to oldItem — only hapOutput was. Calling remove on an output that isn't attached is probably a no-op but it's logically wrong. A self.usingHapOutput: Bool flag, or simply tracking  
  which output is active, would make the teardown correct.
                                                                                                                                                                                                                                                                       
  ---                                                                 
  5. Force unwrap after nil check

  let dxtData = frame.dxtDatas[0]
  guard dxtData != nil else { return nil }
  // ...                                                                                                                                                                                                                                                               
  texture.replace(..., withBytes: dxtData!, ...)
                                                                                                                                                                                                                                                                       
  This should just be guard let dxtData = frame.dxtDatas[0] else { return nil }. The force unwrap is unnecessary and obscures intent.                                                                                                                                  
  
  ---                                                                                                                                                                                                                                                                  
  6. Silent frame drop when allocFrameClosest returns nil             
                                                                                                                                                                                                                                                                       
  guard let frame = hapOutput.allocFrameClosest(to: itemTime) else { return true }
                                                                                                                                                                                                                                                                       
  Returning true claims the HAP path handled the frame, suppressing the standard path — but nothing was emitted. A stall in the Hap decoder means the downstream consumer gets silence with no indication. At minimum this warrants a log; ideally the last good frame 
  should be held.
                                                                                                                                                                                                                                                                       
  ---                                                                 
  7. Loop-back seek ignores seekTolerance
                                                                                                                                                                                                                                                                       
  self.observer = NotificationCenter.default.addObserver(...) { note in
      self.player.seek(to: .zero, toleranceBefore: .zero, toleranceAfter: .zero)                                                                                                                                                                                       
      self.player.play()                                                                                                                                                                                                                                               
  }
                                                                                                                                                                                                                                                                       
  The configured seekTolerance is ignored here — the loop-back always uses exact precision (.zero). For a performance node set up with loose tolerance for fast seeking, the loop-back will be slower than necessary. Should use the same tolerance as performSeek.    
   
  ---                                                                                                                                                                                                                                                                  
  8. Strong self capture in the loop observer                         
                                                                                                                                                                                                                                                                       
  { note in
      self.player.seek(...)                                                                                                                                                                                                                                            
      self.player.play()                                              
  }

  Should be [weak self] note in with a guard let self. The observer is removed on next load so the retain window is bounded, but it's still an avoidable retain cycle during that window.                                                                              
   
  ---                                                                                                                                                                                                                                                                  
  9. srcStride from rgbDataSize / height is fragile                   
                                                   
  let srcStride = frame.rgbDataSize / height
                                                                                                                                                                                                                                                                       
  This assumes the RGB data is padded to exactly rgbDataSize / height bytes per row with no additional alignment. If HapDecoderFrame has a rgbBytesPerRow or similar property, that's what belongs here. If not, this is a silent bug waiting for an odd-width frame.  
                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                       ```

**Comment by tobyspark** [comment]
Actually I’ve got a commit or so more blocked by working on Math Expression. Watch this space.

File: Fabric/Nodes/Image/MovieProviderNode.swift (hunk 11 above)
Comment (tobyspark) [reply Fabric/Nodes/Image/MovieProviderNode.swift]: I hadn’t seen this when I posted below I’ve got a commit or so more. Which addresses some of that.

File: Fabric/Nodes/Image/MovieProviderNode.swift [hunk 13]
```diff
@@ -31,7 +35,8 @@ private let MovieProviderNodeInitializer: Void = {
 
 public class MovieProviderNode : Node, NodeFileLoadingProtocol
 {
-  
+    fileprivate static let log = Logger(subsystem: "graphics.fabric", category: "MovieProviderNode")
+
```
vade: Oh nice, ive never actually used os.logger this looks super useful.

PR #256, opened 2026-05-17
Math Expression: debounced eval + coalesced undo

vade:
Stop re-parsing on every keystroke. Typing "sin(x" used to add the `x` port, fail to parse the next half-formed expression, drop ports, re-add them — every character. Now:

- The TextField writes to a local @State buffer instead of the node's expression, so the AppKit field never gets a re-write of its own text and the cursor / selection survives the eval. (Direct binding to an @Observable property rebuilds the binding on each keystroke, which the macOS TextField representable treats as an external change and pushes back into NSTextField — whose stringValue setter selects-all by default.)
- Each buffer change reschedules a 1s debounce task. Trailing edit fires `commit()`. Enter / defocus pre-empts the debounce and commits immediately. The buffer is the source of truth until then; `node.stringExpression` is left alone so external observers see the user's intended state, not every intermediate keystroke.
- `commit()` registers a single coalesced undo step capturing the full edit range. The symmetric undo→redo pattern bounces correctly on repeated ⌘Z / ⌘⇧Z; both sides re-run `commitExpression()` so the variable ports follow the restored expression.
- External edits (undo / redo / scripted) sync the buffer via `.onChange(of: node.stringExpression)` when the field isn't focused; while the user is typing the buffer stays authoritative.

Node-side: `stringExpression` is now a plain `@Observable` `var` (no `@ObservationIgnored`, no `didSet`) so the View can `.onChange` it. Eval is invoked explicitly via the new public `commitExpression()` from the View, the decoder, and the convenience initialiser — same call sites as before, just spelled out instead of riding `didSet`.

### DIFF
--- a/Fabric/Nodes/Parameters/Number/MathExpressionNode.swift
+++ b/Fabric/Nodes/Parameters/Number/MathExpressionNode.swift
@@ -14,20 +14,102 @@ internal import MathParser
 struct MathExpressionView : View
 {
     @Bindable var node:MathExpressionNode
-    
+
+    /// Local edit buffer between the TextField and the node.
+    ///
+    /// Why this exists: binding the TextField directly to an
+    /// `@Observable` `node.stringExpression` causes the field to
+    /// select-all-on-update on every keystroke. Each write
+    /// invalidates the body, `@Bindable` synthesises a fresh
+    /// `Binding<String>`, and SwiftUI's macOS TextField representable
+    /// reads that as an external change — pushing the value back
+    /// into the underlying `NSTextField`, whose `stringValue` setter
+    /// selects-all by default. With `@State` here, SwiftUI's
+    /// value-equality optimisation kicks in (the value the field just
+    /// emitted equals the value the binding now reports) and the
+    /// AppKit field is left untouched. The cursor / selection
+    /// survives the eval.
+    @State private var buffer: String = ""
+    /// Tracks the in-flight debounce. Cancelled on each keystroke so
+    /// only the trailing edit triggers a re-parse + port rebuild.
+    @State private var debounceTask: Task<Void, Never>?
+    @FocusState private var focused: Bool
+
+    /// How long the field has to stay quiet before the expression
+    /// auto-commits. Enter and defocus bypass this and commit
+    /// immediately.
+    private static let debounceSeconds: Double = 1.0
+
     var body: some View
     {
         VStack(alignment: .leading)
         {
             Text("By writing a mathematical expression, you can expose variables and use built in functions or constants to compute a single output value. \n\n [Swift-Math-Expression Documentation](https://github.com/bradhowes/swift-math-parser).")
-            
+
             Spacer()
-            
-            TextField("Math Expression", text: $node.stringExpression)
+
+            TextField("Math Expression", text: $buffer)
                 .lineLimit(1)
                 .font(.system(size: 10))
                 .textFieldStyle(RoundedBorderTextFieldStyle())
+                .focused($focused)
+                .onAppear { buffer = node.stringExpression }
+                .onSubmit { commit() }
+                .onChange(of: focused) { _, isFocused in
+                    if !isFocused { commit() }
+                }
+                .onChange(of: buffer) { _, _ in
+                    // Schedule the eval. Until it fires (or Enter /
+                    // defocus pre-empts), the buffer is the source of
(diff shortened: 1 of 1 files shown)


vade:
Im not gonna merge this @tobyspark - I *think* theres a better way to to debounce via something like https://medium.com/@anselmus.pavel/debouncing-user-input-in-swiftui-10dda5231bdf

```
lass ResourceViewModel: ObservableObject {
	@Published var selectedResourceTitle = ""	//to store real-time value
	@Published var debouncedTitle = ""		//to store debounced value
	
	init() {
		setupTitleDebounce()
	}
	
	func setupTitleDebounce() {
		debouncedTitle = self.selectedResourceTitle
		$selectedResourceTitle
			.debounce(for: .seconds(0.75), scheduler: RunLoop.main)
			.assign(to: &$debouncedTitle)
	}
```

mind taking a peek?

tobyspark:
I really hope there’s a better way! As flagged in a different channel, this is an improvement but not yet fit for purpose. 

Yes, I’ll look into it.

tobyspark:
I’ve looked at that, and sadly, nope.

The debounce in that example is non-interactive, it’s rate-limiting a back-end function. The UX we are dealing with is interactive: the user does something and expects to see a change. That change can be problematic, as typing can churn ports, necessarily disconnecting etc., and we don’t want undo scoped to keystrokes, we want it scoped to something impacting the mathematical parsing of the string.

A user that has chosen to use the Math Expression node likely knows what they want. So I think the primary expectation for a user here is no problematic mid-user-action ports and undo updates. Rather, have them happen in direct response to signalling the end of the action via enter or defocus. Debounce has a user benefit here too, to cut down on port churn, but that’s a nice to have rather than HIG-esque see the direct response of your action.

Before even getting to implementation details, I think a simple debounce is not fit for purpose here. Plus, I tried it – I started with that kind of debounce implementation based on your comment in the settings thread, and I didn’t like it – having to pause to see the effect decide to kick in just didn't feel “pro” to me. I hit enter instinctively on that field.

The implementation here is more complicated because we have a user in the loop bringing a more complex set of expectations. Now did I land on the best implementation in my pre-gig sprint? No, and I’ll push a commit fixing that  – adopting SwiftUI `.task` and a more idiomatic undo handler.

The “still not fit for purpose” comments I’ve made I have now tracked down to two a Fabric level issues. 

1. The Math Expression node was successfully committing the expression, as e.g. seen in the node title, but in some cases ports were not updating, requiring a click on the node itself (and only there, not the wider graph) to update. Changing the name of a single variable will trigger this, e.g. change `x` to `y` and you will see y in the title but x remain as a port.
2. Clicking on the graph targets key presses to node selection. Clicking back on the settings pane does not target the arrow keys back.

Fixes for these are also following, but they can be treated as an initial suggestion. There are architectural alternatives, etc.

vade:
there seem to be two prs which effect math expression now?

Fabric/Nodes/Parameters/Number/MathExpressionNode.swift
```diff
@@ -14,21 +14,103 @@ internal import MathParser
...
+    /// into the underlying `NSTextField`, whose `stringValue` setter
+    /// selects-all by default. With `@State` here, SwiftUI's
+    /// value-equality optimisation kicks in (the value the field just
+    /// emitted equals the value the binding now reports) and the
+    /// AppKit field is left untouched. The cursor / selection
+    /// survives the eval.
+    @State private var buffer: String = ""
+    /// Tracks the in-flight debounce. Cancelled on each keystroke so
+    /// only the trailing edit triggers a re-parse + port rebuild.
+    @State private var debounceTask: Task<Void, Never>?
```
vade, line 35: i dont believe you need the task stored, nor the manual cancellation.

 if you use task(id:buffer) it will track changes and fire on change with cancellation. 

```
Another variant of the task view modifier allows us to observe equitable data and run the async task whenever the data changes. The task lifecycle is still bound to the view lifecycle, but SwiftUI also cancels the ongoing job whenever data changes and creates a new one for the latest data.
```


https://swiftwithmajid.com/2022/06/28/the-power-of-task-view-modifier-in-swiftui/

tobyspark: That is addressed on the PR I made earlier. (And had to make because I didn’t have write-access to this branch)

PR #259, opened 2026-05-17
Node settings improvements

tobyspark:
From #256, here is the follow-on work. I can’t write to that PR, hence this. Documented in [#256](https://github.com/Fabric-Project/Fabric/pull/256#issuecomment-4471106950) and commit messages here.

### DIFF
--- a/Fabric/EditorInputFocus.swift
+++ b/Fabric/EditorInputFocus.swift
@@ -54,3 +54,36 @@ extension EnvironmentValues
         set { self[EditorInputFocusSetterKey.self] = newValue }
     }
 }
+
+extension View
+{
+    /// Apply to any focusable control (typically a TextField) inside
+    /// a node-settings popover. When the control gains focus, pushes
+    /// the editor's input focus to `.nodeSettings` so the GraphCanvas
+    /// key handler yields arrow keys back to the control.
+    ///
+    /// Bundles the `@Environment(\.setEditorInputFocus)` read and the
+    /// focus-change observer so per-field call sites stay one line.
+    /// Safe to apply to fields that already have their own
+    /// `@FocusState` + `.focused(...)` — SwiftUI tolerates multiple
+    /// `.focused` modifiers on the same field, each tracking it.
+    func reclaimsNodeSettingsFocus() -> some View
+    {
+        modifier(ReclaimsNodeSettingsFocusModifier())
+    }
+}
+
+private struct ReclaimsNodeSettingsFocusModifier: ViewModifier
+{
+    @FocusState private var focused: Bool
+    @Environment(\.setEditorInputFocus) private var setEditorInputFocus
+
+    func body(content: Content) -> some View
+    {
+        content
+            .focused($focused)
+            .onChange(of: focused) { _, isFocused in
+                if isFocused { setEditorInputFocus(.nodeSettings) }
+            }
+    }
+}
--- a/Fabric/EditorInputFocus.swift
+++ b/Fabric/EditorInputFocus.swift
@@ -5,7 +5,7 @@
 //  Created by Codex on 3/9/26.
 //
 
-import Foundation
+import SwiftUI
 
 public enum FabricEditorInputFocus: String, Codable, Hashable
 {
@@ -13,4 +13,44 @@ public enum FabricEditorInputFocus: String, Codable, Hashable
     case registry
     case inspector
     case nodeSettings
+}
+
+/// Setter for the editor's input focus, injected via Environment by
+/// GraphCanvas (which owns the binding). Descendant views — in
+/// particular settings popover content — call this to claim arrow
+/// keys when one of their text fields gains focus.
+///
+/// Why a callback in Environment and not `@FocusedValue`: the
+/// settings popover is presented in its own NSWindow on macOS, and
+/// FocusedValues don't bridge across that boundary, so a focused
+/// value set inside the popover isn't visible to GraphCanvas.
+/// SwiftUI environment values DO propagate to popover content.
+///
+/// Defaults to a no-op so call sites don't need optional chaining
+/// when no GraphCanvas ancestor is present (matching SwiftUI's
+/// own `dismiss` / `openURL` pattern).
+///
+/// Applied per text field — any TextField inside a settings view
+/// that should hold the arrow keys reads this from the Environment
+/// and invokes it with `.nodeSettings` when its `@FocusState`
+/// becomes true.
+///
+/// Note: the Environment value is a bare closure. Closures aren't
+/// Equatable, so each GraphCanvas body re-eval injects a "new"
+/// closure and SwiftUI invalidates descendant readers. Tolerable
+/// here (GraphCanvas re-evals infrequently, popover content is
+/// cheap). If more callers appear, consider wrapping in a
+/// `DismissAction`-style struct.
+private struct EditorInputFocusSetterKey: EnvironmentKey
+{
+    static let defaultValue: @MainActor (FabricEditorInputFocus) -> Void = { _ in }
+}
+
+extension EnvironmentValues
+{
+    var setEditorInputFocus: @MainActor (FabricEditorInputFocus) -> Void
+    {
+        get { self[EditorInputFocusSetterKey.self] }
+        set { self[EditorInputFocusSetterKey.self] = newValue }
+    }
 }
--- a/Fabric/Graph/Port/NodePort.swift
+++ b/Fabric/Graph/Port/NodePort.swift
@@ -161,7 +161,7 @@ public class NodePort<Value : PortValueRepresentable>: Port
             port.connect(to: other)
         }
         self.node?.graph?.undoManager?.setActionName("Disconnect Ports")
-        self.node?.graph?.shouldUpdateConnections.toggle()
+        self.node?.graph?.shouldUpdateConnections = true
     }
 
     override public func connect(to other: Port)
@@ -256,7 +256,7 @@ public class NodePort<Value : PortValueRepresentable>: Port
             port.disconnect(from: other)
         }
         self.node?.graph?.undoManager?.setActionName("Connect Ports")
-        self.node?.graph?.shouldUpdateConnections.toggle()
+        self.node?.graph?.shouldUpdateConnections = true
     }
 
     public func send(_ v: Value?, force:Bool = false)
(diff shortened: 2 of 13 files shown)


Fabric/Graph/Port/PortRegistry.swift
```diff
@@ -24,9 +24,18 @@ import Foundation
   resorting to Objective-C runtime features or unsafe reflection.
  */
 
+@Observable
```
vade, line 27: I dont think @Observable with private variables does anything? unless you want ordered to be specifically surfaced as `public private(set)` ?

Fabric/Graph/Port/NodePort.swift
```diff
@@ -250,21 +250,13 @@ public class NodePort<Value : PortValueRepresentable>: Port
...
-            self.send(self.value, force: true)
-        }
+        self.send(self.value, force: true)
 
         self.node?.graph?.undoManager?.registerUndo(withTarget: self) { port in
             port.disconnect(from: other)
         }
         self.node?.graph?.undoManager?.setActionName("Connect Ports")
-        self.node?.graph?.shouldUpdateConnections.toggle()
+        self.node?.graph?.shouldUpdateConnections = true
```
vade, line 259: Nice catch(es) on this failure!

Fabric/EditorInputFocus.swift
```diff
@@ -14,3 +14,76 @@ public enum FabricEditorInputFocus: String, Codable, Hashable
...
+/// and invokes it with `.nodeSettings` when its `@FocusState`
+/// becomes true.
+///
+/// Note: the Environment value is a bare closure. Closures aren't
+/// Equatable, so each GraphCanvas body re-eval injects a "new"
+/// closure and SwiftUI invalidates descendant readers. Tolerable
+/// here (GraphCanvas re-evals infrequently, popover content is
+/// cheap). If more callers appear, consider wrapping in a
+/// `DismissAction`-style struct.
+private struct EditorInputFocusSetterKey: EnvironmentKey
```
vade, line 44: see my main comment / concerns from claude review of this approach

vade (commented):
PR #259 — Performance Review: EditorInputFocus.swift & related                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                       
 Architectural choice: environment closure vs. FocusedValue                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                       
  The rationale is correct. FocusedValue doesn't cross NSWindow boundaries on macOS; SwiftUI environment values do. Given the popover-in-own-NSWindow constraint, this is the right architectural call.                                                                
                                                                                                                                                                                                                                                                       
  ---                                                                                                                                                                                                                                                                  
  Primary concern: closure equality + self-reinforcing invalidation   
                                                                                                                                                                                                                                                                       
  The author flagged this themselves in the comment but incorrectly rated it as "tolerable":
                                                                                                                                                                                                                                                                       
  // Note: the Environment value is a bare closure. Closures aren't   
  // Equatable, so each GraphCanvas body re-eval injects a "new"                                                                                                                                                                                                       
  // closure and SwiftUI invalidates descendant readers. Tolerable                                                                                                                                                                                                     
  // here (GraphCanvas re-evals infrequently, popover content is cheap).                                                                                                                                                                                               
                                                                                                                                                                                                                                                                       
  The claim that GraphCanvas re-evals infrequently is wrong in this context. GraphCanvas.body re-evals any time inputFocus changes — and inputFocus changes precisely when a user focuses a TextField, which calls setEditorInputFocus(.nodeSettings), which mutates   
  inputFocus. This creates a self-reinforcing loop:                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                       
  TextField gains focus                                               
    → setEditorInputFocus(.nodeSettings) called                                                                                                                                                                                                                        
      → GraphCanvas.inputFocus mutates
        → GraphCanvas.body re-evals                                                                                                                                                                                                                                    
          → new closure injected into environment                                                                                                                                                                                                                      
            → all \.setEditorInputFocus readers invalidated
              → ReclaimsNodeSettingsFocusModifier.body re-evals                                                                                                                                                                                                        
                                                                      
  SwiftUI prevents infinite cycling (focused state didn't change, so .onChange doesn't re-fire), but the unnecessary re-eval of every \.setEditorInputFocus reader happens on every focus event. Additionally shouldUpdateConnections = true is now set on             
  connect/disconnect/rename operations, which also drive GraphCanvas re-evals.
                                                                                                                                                                                                                                                                       
  Your intermediate-view concern is partially mitigated by SwiftUI's dependency tracking — views that don't read @Environment(\.setEditorInputFocus) are not re-evaluated just because an ancestor injects a new value. However, all views that do read it are         
  unnecessarily invalidated, including anything using .reclaimsNodeSettingsFocus().
                                                                                                                                                                                                                                                                       
  Fix: wrap in an always-equal struct (DismissAction pattern)                                                                                                                                                                                                          
  
  The author mentioned this but deferred it. It should be applied now. The pattern is exactly how SwiftUI implements DismissAction, OpenURLAction, etc.:                                                                                                               
                                                                      
  // EditorInputFocus.swift                                                                                                                                                                                                                                            
  public struct EditorInputFocusSetter {                                                                                                                                                                                                                               
      fileprivate let action: @MainActor (FabricEditorInputFocus) -> Void
      func callAsFunction(_ focus: FabricEditorInputFocus) { action(focus) }                                                                                                                                                                                           
  }                                                                                                                                                                                                                                                                    
  
  extension EditorInputFocusSetter: Equatable {                                                                                                                                                                                                                        
      // Always equal: GraphCanvas is the permanent owner and the action
      // always targets the same instance. Prevents descendant invalidation
      // on every GraphCanvas body re-eval.                                                                                                                                                                                                                            
      static func == (lhs: Self, rhs: Self) -> Bool { true }
  }                                                                                                                                                                                                                                                                    
                                                                      
  private struct EditorInputFocusSetterKey: EnvironmentKey {
      static let defaultValue = EditorInputFocusSetter { _ in }
  }

  extension EnvironmentValues {
      var setEditorInputFocus: EditorInputFocusSetter {
          get { self[EditorInputFocusSetterKey.self] }                                                                                                                                                                                                                 
          set { self[EditorInputFocusSetterKey.self] = newValue }
      }                                                                                                                                                                                                                                                                
  }                                                                   
                                                                                                                                                                                                                                                                       
  In GraphCanvas.swift, use a weak capture to avoid a retain cycle:                                                                                                                                                                                                    
  
  .environment(\.setEditorInputFocus, EditorInputFocusSetter { [weak self] focus in                                                                                                                                                                                    
      self?.inputFocus = focus                                        
  })

  Call sites don't change — EditorInputFocusSetter with callAsFunction means you still call it as setEditorInputFocus(.nodeSettings).                                                                                                                                  
  
  ---                                                                                                                                                                                                                                                                  
  Secondary concern: ReclaimsNodeSettingsFocusModifier creates a second @FocusState
                                                                                                                                                                                                                                                                       
  Every use of .reclaimsNodeSettingsFocus() adds an independent @FocusState private var focused: Bool inside the modifier, alongside whatever @FocusState the call site already has. The comment says "SwiftUI tolerates multiple .focused modifiers" — this is true,
  but it means:                                                                                                                                                                                                                                                        
                                                                      
  - Two separate FocusState subscribers tracking the same field                                                                                                                                                                                                        
  - On focus change, both fire .onChange(of:) handlers independently  
  - If a call site uses the modifier AND has its own @FocusState for commit-on-defocus (like MathExpressionView), the modifier's internal state is redundant                                                                                                           
                                                                                                                                                                                                                                                                       
  MathExpressionView is correctly not using .reclaimsNodeSettingsFocus() — it handles focus directly and inline because it needs the defocus event for commit(). That's the right call. The other nodes (OSCReceive, Timeline, StringFormatter, StringScanner) apply it
   to TextFields without their own FocusState, which is clean.                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                       
  ---                                                                 
  The other changes in this PR

  ┌──────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                        Change                        │                                                                                             Assessment                                                                                              │
  ├──────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤       
  │ shouldUpdateConnections.toggle() → = true everywhere │ Correct. Toggle was a latent bug where paired ops (rename = remove + add) cancelled silently.                                                                                                       │
  ├──────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤       
  │ @Observable on PortRegistry                          │ Correct. Automatic observation replaces the brittle toggle pattern. More fine-grained than the old .id(shouldUpdateConnections) rebuild.                                                            │       
  ├──────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤       
  │ MathExpressionView buffer + .task(id:) debounce      │ Sound. .task(id:) cancellation is the idiomatic debounce. The buffer-breaks-AppKit-select-all explanation is accurate. Undo coalescing (all keystrokes since last commit → one ⌘Z) is the right UX. │       
  └──────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘       
                                                                      
  ---                                                                                                                                                                                                                                                                  
  Summary                                                             
                                                                                                                                                                                                                                                                       
  The architecture is correct. The one real issue is the bare-closure environment value — the author's own analysis identified the problem but underestimated how often GraphCanvas re-evals in the focus path. Apply the always-equal struct wrapper now. Everything
  else in the PR is solid.

Fabric/Graph/Port/NodePort.swift
tobyspark: Nothing like using something in anger.

vade:
Hey so, due to the changes with NodeViewModel etc, i think its worth revisiting some of this 

Im going to pull out 'safe' changes like the toggle vs true related things - but some of the larger focus / settings things now might be more easily expressed due to node / model changes. 

Im thinking that we refactor Graph into a two or three layer view, one that has Notes, Nodes, and Settings (maybe maybe even patch coords come to think of it) which would allow for node ports to update without retriggering the Settings popover animation. 

That doesnt address input focus, but i get a sense you might have more of a handle on that than I?

vade:
Given the latest PR merge i think this can be closed?

PR #262, opened 2026-06-08
Fix for MIDI Input manufacturer handling

dpwolf:
Before this fix I wasn't able to build the project due to the following error:

```
.../Fabric/Fabric/Nodes/Parameters/IO/MIDIInputNode.swift:469:40 Cannot convert value of type 'MIDIIOObjectProperty.Value<String>' to expected argument type 'String'
```

Let me know if the tests are overkill.


vade:
weird, what are you building against? Fabric builds fine for me without this change? What OS / SDK / XCode?

dpwolf:
I'm on Tahoe 26.5.1 (25F80), Xcode 26.5, base SDK macOS.

vade:
The issue was a changed version in Swift PM which silently upped a version which had a API change. - ive changed the pinned version

PR #264, opened 2026-06-15
Feature – Array node expansion, including Array Math Expression and procedural layout nodes

tobyspark:
## New nodes

- **Array Math Expression** — expression evaluated once per output element
  (`i`/`n`/`t` bindings, dynamic variable ports).
- **Procedural layout** — Line / Ring / Grid Points; **Transform Array From TRS**
  (zips Positions/Orientations/Scales).
- **Vector ⇄ components** (scalar + array, both directions):
  - compose: `Vector N From XY…` (scalar) · `Vector N Array From XY…` (array)
  - split: `XY… From Vector N` (scalar) · `XY… Array From Vector N` (array)
- **Fill** — `… Array With Count` (replicate) · `… Array From Rippled Value`
  (time-windowed resample).
- **Orientation** (scalar + array, symmetric) — `Orientation [Array] From Euler /
  Axis Angle / Target / Tween` (the Target node is the collapsed look-at;
  the array tween shares one clock across elements).
- **Vector Tween** (scalar + array) — `Vector 2/3/4 [Array] Tween`.

## Shared infrastructure

- `MathExpressionBaseNode` (base for the two math-expression nodes).
- **`TweenDriver<Value>` / `TweenArrayDriver<Value>`** — one tween state machine,
  parameterized by an interpolation closure; used by the Vector and Orientation
  tweens (lerp vs slerp).
- `ContiguousArray.paddedToLast`; `simd_quatf` extensions (`safeVector`,
  `lookingAlong:up:`, `eulerAnglesDegrees:`, `axis:angleDegrees:`,
  `upDirection(from:)`); `makeInspectorValuePort` (Float / Vector2/3/4
  inspector-editable generic inlets).

## Optimizations & fixes

- Per-element hot-loop reductions (constant-rotation hoist; T·R·S closed form).
- Line/Ring spacing made endpoint-accurate; Ring gained a **Closed/Open** toggle.
- NaN/Inf and degenerate-quaternion guards; `MakeQuaternionNode`
  axis-normalization bug fixed; `makeInspectorValuePort` Float4 default
  corrected to `.zero`.

## Naming / consistency

- Adopted **`<output> From <input>`** across the array/vector conversion nodes;
  `… From Value` → `… With Count`; `… From Values` → `… From XYZ`.
- Aligned branch-new **class/file names** to display names; folded the registry's
  Quaternion section into Vector.

## Examples 

Example of using Array nodes with Math Expressions to create a procedural layout –

Screenshot 2026-06-15 at 23 03 47

Screenshot 2026-06-15 at 23 04 27

Example of using higher-level array nodes to create a procedural layout –

Screenshot 2026-06-15 at 23 30 29

Example of using the ripple fill node and de-/re-composing Vector N arrrays for processing –

Screenshot 2026-06-15 at 23 58 37


tobyspark:
What this doesn’t have
- Array Math Expression with no variable array inputs (from which count would be inferred) and a count input instead. Of the ways to handle output array count, count via supplied arrays has ease for the common case and is consistent with other nodes. Add a constant variable and set via Array With Count.
- Array Math Expression that can operate on Vec2/3/4. Decompose to component arrays.

tobyspark:
[PR264 examples.zip](https://github.com/user-attachments/files/28978478/PR264.examples.zip)

vade:
So quick thoughts so far! 

Theres a lot of functionality in here and useful nodes!

There are some redundant nodes that can be made using `GeometryToTransformArrayNode` without needing to output just points, or at a minimum we can update `GeometryToTransformArrayNode` to output points / normals / transforms (transforms are helpful for instanced rendering of one geo with another)

* Line Points
* Grid Points
* Ring Points

I think those can go in lieu of using `GeometryToTransformArrayNode` or an update / alternative `GeometryToPositionArrayNode` or maybe a `GeometryToDataNode` that outputs positions, normals and what not as appropriate arrays. 

Code wise -  I feel like theres redundancy in the code base and using Generics / Templating could make this more succinct and offer a nice systems approach to getting components out, and adding components together, but i need to see if thats realistic.  There's also some Satin conventions we should adopt here, like Satin.worldUpDirection / worldForwardDirection, worldRightDirection.

3) Naming and conventions - this is where things get tricky, but i think we should be strategic with how we allow for things to get built and name / group them accordingly. Let me know if this makes sense? 

For all Numeric Arrays I think we want options like the following:

* Array Creation - 
   * Queue
   * Repeat 
   * Interpolate (start and end values + easing / and final count)
   * Expression (provide a math expression thats evaluated n (count) times by passing the index in - ie sin(i) or some such)
   * Sub array / window - make a smaller array from a portion of a larger array
    * Array Combine - append arrays with the same type together
    
* Arrays Composition - ie i want to make an array of Vector 2s from 2 equal length arrays of Floats. 
      * Arrays of Floats -> Array of Vector 2, 3, 4, Orientation
      * Arrays of Vectors -> Transform

* Array Decomposition - ie i want to turn an array of Vector 2's into 2 Arrays of Floats. 
   * Reverse of above

* Array Wrapping - I want to turn a long Array of Floats into a smaller array of Vector 2/3/4/ or even Transforms. 

* Array Upcasting - I want to take an array of lower dimension and turn it into an array of higher dimension. ie Float to Vector 2 - using a default value, repeating the value.

* Array Math - I have n existing values and I want to do things with it
    * Apply Expression (for numerics but not vector or matrix, at least, i dont think our math expression supports that)
    * Apply Transform (apply a single transform to all elements - for Vector 4 Array and Transform Array))
    * Apply Transforms (apply an array of transforms to all elements - for Vector 4 Array and Transform Array)

* Array Interpolation - Change count of items by interpolating intermediate values
* Array Decimation - Reduce count of values
* Array Tweening - 
    * Produce one array from 2 arrays with 2 sets of values (ie tween between two array of float 3 positions).
    

LMK your thoughts, i think this is close and we can identify gaps, and some naming conventions and potentially adopt generics for some of this 

Thanks this is gonna be dope!

vade:
Maybe the thing to do is 

* just look for some implementation issues now
* if you agree with the general premise of the above, we can turn it into a issue / feature  for fully fledged Numeric Array stuff and work through any changes we need to make.

tobyspark:
Using geometry nodes to derive arrays of Transforms and/or TRS would be good composability. Currently fails at the first thing I wanted to do months back which was a line of points, i.e. it’s amazing you can get a point cloud of an arbitrary 3D model but we don’t want to miss or over-complicate the basics. e.g. where’s the line primitive? e.g. how do I set the count on `Circle Geometry`, the only input is ‘Size’.

Prompted by the above, concretely this PR is missing `TRS Array from Transforms`. Though performance-wise we’d likely want `TRS Array From Geometry` anyway, I assume.

tobyspark:
In principle yes to Fabric having a “feature complete” set of Array nodes spanning scalars/vecs/matrices. Seems a no brainer, but I do have a nagging thought that we might be missing a trick. I think the root of that comes from my experience with QC where moderately complicated data processing was easier to express and faster in a single Javascript node vs. a spaghetti of single-function nodes. 

So before going full-bore on the permutations of array nodes, two “modern” thoughts 
- Not being able to express vector calculations in the Array Math Expression node, and not being able to intrinsically express count when it’s purely a generative (array creation vs. processing), makes me wonder what there might be that could. e.g. something like a MATLAB node?
- Point Clouds, and doing stuff with them, should stay on the GPU. This isn’t that. What might be?

tobyspark:
From conversation we are likely going to merge this before attempting to collapse the number of variants; #269 will likely surface candidates Fabric-wide beyond the nodes here.

The Line / Ring / Grid Points nodes remain questionable. I think composability wins so my preference would be do #268 externally to this PR, remove the Points nodes from this PR, and merge.

tobyspark:
Points From Geometry / Geometry From Points

Screenshot 2026-06-17 at 13 08 10

Existing behaviour (with improved naming)

Screenshot 2026-06-17 at 12 58 09

Screenshot 2026-06-17 at 12 59 50

New behaviour

Screenshot 2026-06-17 at 13 05 21

Screenshot 2026-06-17 at 13 05 26

vade:
HELL YES this is great.

PR #274, opened 2026-06-19
Feature/satin 2.0 node model view

vade:
This PR removed https://github.com/observable from our Node model, and creates a shim NodeModelView which is responsible for view based data and is what SwiftUI interacts with.

This removes some properties and separates responsibilities from Node to NodeModelView. Framesing, updates, UI etc all migrate to NodeModelView.

For nodes with Settings, we migrate settings to a SettingsModel to bridge.

This is a step to help optimize SwiftUI / Canvas and also to isolate concerns such that once Satin gets merged in threaded rendering we can isolate calls easily from main thread / updates to rendering thread.


vade:
Whoops meant to merge this into Feature / 2.0 - derp.

PR #278, opened 2026-06-20
Feature/satin 2.0 port type fixes

vade:
This is a daft PR exploring

* updating the type system to support virtual types

* Introducing Typed nodes, ie removing specialization in the node registry, and having Single "Array" nodes that handle all possible types, including Virtual (only one Array Queue node for example)

* Updating the UI and internals to have the above work (can connect)

* centralizing some important code that was sort of spread around that made the above hard. 

* introducing a new `TypeAgnosticNode` that builds off of the 'Strategy" node subclass that handles the common type cases in a standard interface. 

* thinking about specialization for specific groups of types, like numeric parameters (ie one Compose Vector Node) that lets you chose Vector 2, 3, or 4 for example.


vade:
@tobyspark def want your opinion on this, its working at semi firs glance, but this is brand new and ive yet to get super into it. General idea is here tho!

tobyspark:
This is good.

Virtual typed nodes have some compelling uses. The simplest example: `Array Count` is just that, and it just works.

Typed nodes have some compelling uses, e.g. a single `Vector Compose` node changing the number of inputs accordingly.

Open questions in my mind having used it a bit
- In a world with e.g. a typed `Tween` node (e.g. float, orientation, colour) should there be a separate `Array Tween` node with array typed equivalents?
- Compose/Decompose: we have separate nodes for some types (i.e. vector vs orientation), and then some of those nodes are typed themselves (ie. vec2,3,4 within vector). We could have a single compose and a single decompose node with a flat list of all the available types. Plus this list could collapse the single vs. array nodes.
- There is no good UX reason to expose the type on `Array Count`. I know why it’s there, but that raises a systems-design flag to me. The two examples I opened with are clear example each of where virtual is appropriate (but typed not) and typed appropriate (but virtual not). There is going to be some fuzziness in the node space and we should be thoughtful.

Some UX snafus to clear
- If you wire up something to a virtual typed node, and change to the correct type, the wire disconnects unnecessarily.
- Consistency in node titles. Do they just keep their listing name, or do they gain the typed specialisation? Currently there’s a mix, and e.g. Vector Array Compose doesn’t update its typed title on type change.

vade:
Yea, theres a few nodes that could get partial type treatment. Today, only fully generic nodes (every type expressible) are now a Typed Node with the settings UI.

Tomorrow, partial typed nodes (numerics etc) should get typed node treatment (single node, settings, menu). This is not yet finished. Families of functionality should get combined

We also want to revisit once Color / Signal and maybe Orientation get fully qualified types vs proxy type. 


* Repeat Value- this should actually be fully generic 
* Ripple Repeat Value - i'll be honest the phrasing of the title/ description confuses me but whatever if its useful great!

* Array Resample - Numeric Types and Color
* Interpolate - Numeric Types and Color
* Tween - Numeric Types and Color
* Distance - Numeric types and Color (Cie Delta E 2000 for example)

> Array Tween

Yea, I dont see why not.  and Ease (aside:  one is manual lerpinng functions - other is input time with looping i guess - we should semantically be clear - we work differently and i lean on raw easing funcs with normalized input more)

> Compose

I suppose a single compose / decompose could work? Not sure why i can't conjugate hesitation, its probably a good idea :) 

> Array Count

Thats a fair point, but theres small performance costs to virtual arrays vs type arrays even in message passing. I dont mind it having a pull down. 

UI Nits:

 > If you wire up something to a virtual typed node, and change to the correct type, the wire disconnects unnecessarily. 
 
 Sure, nice to have if and only if the underlying ports are compatible. We can try a reconnect?
 
 
> Titles

Should get simpler given the first notes about combining partual types, as this is currently very WIP. 

I think it would be good to be able to 'search' for Vector 4 Tween and get it 'initted right' but thats a beauty nice to have pass - i think it would just be 'Tween' for now.

vade:
Oh I also think we may want to re-organize the insepctor search / sidebar  with a Animation / Signal group, and a Array group once we get there.

tobyspark:
Could a virtual type node set its type on first connection to it? Still have the drop-down for subsequent use, but that would go a long way for the UX here, right?

vade:
Maybe - but I suspect thats a non trivial change for some subtle reasons. 

My interim goal for this PR is to get the type / messaging system (and thus execution - see iterator bs) working well and 'settled' behaviour wise, and then we can add some UX and optimization passes on top in separate PRs.

vade:
This has soooo much stuff going on, going to merge before it gets further out of hand!

PR #283, opened 2026-06-25
Split Array Math Expression into separate Processor and Generator nodes

tobyspark:
In the original Array Math Expression PR I called out there was a design trade-off around having a `Count` input. I decided a _processor_ idiom was the better choice, where in the common case an input array is processed element-wise to the output.   This PR says “why choose”, as a _generator_ idiom is also useful and we should support it trivially. Scales from filling an array with a constant, stepping through a range, up to whatever you can do with the supported math functions and the `i`, `n`, and `t` per-element bindings.

> Array Math Processor
> _Use a mathematical expression to process numeric array input(s) into an output array. The output length follows the longest input; shorter inputs pad with their last element. Built-in bindings i (index), n (count) and t (progress 0 to 1) are available per element._

> Array Math Generator
> _Use a mathematical expression to generate a numerical array based on count and any constant inputs. Built-in bindings i (index), n (count) and t (position 0 to 1) are available per element._


vade:
Is this still pertinent with the new MathEngineNode?

tobyspark:
Superceded by the far better #285

PR #285, opened 2026-07-11
Feature – Math Expression node uses new MathExpressionEngine

tobyspark:
A far more capable and expressive Math Expression node, collapsing the need for Array Processor / Generator variants, and much graph complexity. Achieved by replacing the [swift-math-parser](https://github.com/bradhowes/swift-math-parser) dependency with a custom, designed-for-Fabric, [MathExpressionEngine](https://github.com/tobyspark/MathExpressionEngine). See [usage guide](https://github.com/tobyspark/MathExpressionEngine/blob/main/GUIDE.md), as linked from the node’s settings.

Screenshot 2026-07-11 at 15 13 58

Screenshot 2026-07-11 at 15 14 24

Scales beyond single-line, single-output expressions with a multi-line editor complete with inline warnings etc.

Screenshot 2026-07-11 at 15 17 05

Note while existing expression strings are compatible with the new engine, the node's default output port has changed from `number` to `result` as the output can be any numeric type, i.e. vec2/3/4, transform, array variants.

The MathExpressionEngine is currently points at the development repository owned by tobyspark. This PR should see that forked into the Fabric organisation and the package link updated.


tobyspark:
Note there’s some keyboarding issues that need resolving, which IIRC is a more general settings-pane issue that also IIRC need to (re-)address after the refactor of graph drawing / layout.

PR #288, opened 2026-07-13
Fixes – Misc

tobyspark:
Various fixes encountered in some reworking of *spark stage and general use of Fabric. Commit messages aim to be sufficiently descriptive, regression tests excise the encountered issue.

### DIFF
--- a/Fabric/Graph/Port/NodePort.swift
+++ b/Fabric/Graph/Port/NodePort.swift
@@ -146,7 +146,7 @@ public class NodePort<Value : PortValueRepresentable>: Port
         }
         else
         {
-            while let index = other.connections.firstIndex(where: { $0.id == other.id } )
+            while let index = other.connections.firstIndex(where: { $0.id == self.id } )
             {
                 other.connections.remove(at: index)
             }
--- a/Fabric/Graph/Port/NodePort.swift
+++ b/Fabric/Graph/Port/NodePort.swift
@@ -212,7 +212,19 @@ public class NodePort<Value : PortValueRepresentable>: Port
         {
             return
         }
-        
+
+        // Connecting an already-connected pair is a no-op. Graph decoding
+        // restores connections from a map keyed by both endpoints, so each
+        // connection gets connected twice; without this guard the second call
+        // would disconnect the pair first, force-sending nil into the inlet
+        // and destroying the freshly decoded parameter value.
+        if self.connections.contains(where: { $0.id == other.id }),
+           other.connections.contains(where: { $0.id == self.id })
+        {
+            return
+        }
+
+
         if self.kind == .Inlet && other.kind == .Outlet
         {
             self.connections.forEach { [weak self] in
--- a/Fabric/Nodes/NodeRegistry.swift
+++ b/Fabric/Nodes/NodeRegistry.swift
@@ -10,9 +10,24 @@ import Satin
 import UniformTypeIdentifiers
 
 public class NodeRegistry {
-    
+
     public static let shared = NodeRegistry()
-    
+
+    init() {
+        // Swift `lazy var` initialization is not thread-safe. Force-build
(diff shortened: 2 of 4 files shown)


tobyspark:
Ah, I have to fork and clone Satin to contribute, so the submodule bump in this PR isn’t going to be correct.

tobyspark:
https://github.com/Fabric-Project/Satin/pull/48

Fabric/Graph/Port/NodePort.swift
```diff
@@ -146,7 +146,7 @@ public class NodePort<Value : PortValueRepresentable>: Port
         }
         else
         {
-            while let index = other.connections.firstIndex(where: { $0.id == other.id } )
+            while let index = other.connections.firstIndex(where: { $0.id == self.id } )
```
vade, line 149: oh man what a lame bug haha

Fabric/Graph/Port/NodePort.swift
```diff
@@ -212,7 +212,19 @@ public class NodePort<Value : PortValueRepresentable>: Port
...
             return
         }
-        
+
+        // Connecting an already-connected pair is a no-op. Graph decoding
+        // restores connections from a map keyed by both endpoints, so each
+        // connection gets connected twice; without this guard the second call
+        // would disconnect the pair first, force-sending nil into the inlet
+        // and destroying the freshly decoded parameter value.
+        if self.connections.contains(where: { $0.id == other.id }),
```
vade, line 221: Nice find.

Fabric/Nodes/Parameters/Array/Array Basics/ArrayCountNode.swift
```diff
@@ -13,7 +13,7 @@ public class ArrayCountNode: TypeAgnosticNode
     public override class var nodeType: Node.NodeType { .Parameter(parameterType: .Array) }
     override public class var nodeExecutionMode: Node.ExecutionMode { .Processor }
     override public class var nodeTimeMode: Node.TimeMode { .None }
-    override public class var nodeDescription: String { "Returns the number of elements in an array. Choose element type in Settings." }
+    override public class var nodeDescription: String { "Returns the number of elements in an array, or 0 if the input is not a countable array. Choose element type in Settings." }
```
vade: Why not nil output ? Semantically its more inline with 'not applicable' than 'this is zero'

Fabric/Nodes/Parameters/Array/Array Basics/ArrayCountNode.swift
```diff
@@ -51,8 +51,18 @@ public class ArrayCountNode: TypeAgnosticNode
...
+        // than being ignored — ignoring it would leave the previous count as a
+        // plausible-looking stale output after a rewire or disconnect.
+        if let boxed = inputPort.snapshotValue(),
+           case .Array(let elements) = boxed
+        {
+            outputPort.send(elements.count)
+        }
+        else
+        {
+            outputPort.send(0)
```
vade: I'd prefer nil myself, why is 0 preferred? node inlets should generally guard on values and not execute at all with nil inputs and port values are nillable for a reason (not just implying no connection)

vade (commented):
One main q - array count of zero vs nil? I think i'd prefer nil as its more 'this doesnt make sense' semantics

tobyspark:
Note the explanation is in the commit message for that. In my words, here –

`nil` is not a valid number. Sending nil means the graph is left with the last valid number. So the user does not receive your intent of signalling “does not make sense”, and worse is left with a likely plausible value and doesn’t even realise what is going on. That happened to me.

Note also, we are arguing about something that there is no actual good answer for, in a situation where the ideal design doesn’t even allow for this class of issue. e.g. I don’t see how it could be controversial that the ideal UX here is the node won’t allow non-array types to connect to it (see discussion on Discord).

PR #294, opened 2026-07-20
Feature/satin 2.0 port tool tip updates

vade:
This pulls in @tobyspark 's changes for Port Previews with values, but simplifies them a bit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,23 +81,29 @@ For all development:
 - Execution is **pull-based**; one execute per node per pass.
 - `GraphRenderer` (executor and scheduler) today does not use `nodeExecutionMode` or `nodeTimeMode` but will in the future.
 - **Iterator (QC-style)** remains the multi-evaluation macro; refinements allowed, paradigm fixed.
+- One file per Node Class
 
 - Node Settings:
   - Nodes may opt into a QC like ’Settings View’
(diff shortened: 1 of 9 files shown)


tobyspark:
I found the lowered character limit lowered the utility significantly. I’ve updated to a 2D heuristic, x lines of y chars. More useful for a broadly similar amount of space. A refinement might be to truncate in a middle line.

Also it seems the current implementation cannot support live updates. I started playing with this but it got in the way of actually getting this merged. According to the genie –

> .help(snapshot) renders a native AppKit tooltip. AppKit reads the tooltip string once, when the tooltip window opens, and never re-reads it. So even though snapshot is @State, updating it while the tooltip is visible changes nothing on screen — the value is frozen at whatever it was when the tooltip appeared. Worse, snapshot is only refreshed on hover-enter, so it's actually frozen at hover-start.
> 
> What would enable it
> 
> You have to stop using the native tooltip and render your own SwiftUI view that re-renders:
> 
> 1. Poll or observe the value. A .task/timer that re-reads port.inspectionTooltip every ~200ms into @State (or make the port @Observable so reads invalidate automatically).
> 2. Render it as a real SwiftUI view, not .help(). A Text(snapshot) in an .overlay/.popover genuinely re-renders when the state changes.
> 3. Escape the node's clip. NodeView clips its content (NodeView.swift:51), so the tooltip must be drawn at the canvas level (like the drag-preview line in GraphConnectionsView), positioned via editingContext.canvasPosition(for:).

Issue #293, opened 2026-07-20
Node names: Type name, Dynamic name, User name. UI: if dynamic or user name present, show type name styled secondary


tobyspark:
Currently nodes have a node-provided name, which can be overridden with a user supplied name. Some nodes dynamically change that provided name. In using Fabric, I came to the view that we should make the dynamically provided name a first-class feature (i.e. part of the base node class). The rationale is that displaying the type of node (i.e. the user-facing type name) is something we don’t want to lose in the UI, and currently the mechanism the node has for making this dynamic overrides that.

The UI design I arrived at – 

<img width="192" height="184" alt="Image" src="https://github.com/user-attachments/assets/c057bded-6f3f-49be-b379-2746ef09ffe8” />

The Movie Provider node has supplied the filename (without extension) as the dynamic name. The Number node is showing the user’s supplied name. Note these short strings are the best case for the design, I started with a two-line design to allow full-width strings but couldn’t get something I was happy with.

[The implementation behind that UI](https://github.com/tobyspark/Fabric/commit/222f78cb7b38449e8cc16d3b7fe5a9ed52fad221) that expressed the intent within `NodeTitleView` keeping the base node class unchanged was not elegant. Hence this issue.

tobyspark:
Fixed in #295

PR #295, opened 2026-07-21
Feature/satin 2.0 node title cherry picks

vade:
Cherry Pick Tobys edits to node titles, and update the inspector to match.


tobyspark:
I’ve implemented #293 which solves a whole class of issues. Also refined the design. 

e.g. As a user you couldn’t set a custom name for the Math Expression node, as it overrode the `displayName` field with the expression. Now we cleanly have the node class name, a name each node instance can generate, and a user supplied name.

This screenshot shows the design refinement. The truncated title with ellipsis was jarring and lost a character or two, now it clips with a fade. The Math Expression node now picks out the salient part of the expression to show...

Screenshot 2026-07-23 at 09 52 19

...which if the first line is a title, will be that.

Screenshot 2026-07-23 at 09 52 10

I also ran with @vade’s inspector tweaks, needed as white titles were near-invisible.

Screenshot 2026-07-23 at 12 53 17

Screenshot 2026-07-23 at 12 52 57

tobyspark:
Ready to merge. Holding off just in case you want to review today's fixes / improvements.

Also note there’s now CLAUDE symlinked riding in this PR _\<shrug>_

Issue #297, opened 2026-07-24
String split to 3D Text


jvcleave:
I was trying to create this chain

<img width="814" height="190" alt="Image" src="https://github.com/user-attachments/assets/6f1df11a-e930-4174-8aa1-d73fd2a457c0" />

but Value wouldn't connect to anything even to Log

I was using `\n` as the delimiter

codex had a fix so I will send that as a PR to see if you want it

vade:
Weird, something must have broke in the interim, that absolutely used to work :X

jvcleave:
I meant to add this to the PR. Here is the summary of what the issue was in case it helps figure out any other issues

## Summary

Fixes text-file line splitting and connections from type-agnostic array nodes to typed inputs such as 3D Text Geometry.

## What was happening

`String Split` treated separators literally, so entering `\n` produced a backslash and `n` instead of a newline.

Dynamic nodes rebuild their ports when their selected type changes. During that rebuild, the canvas could briefly retain the retired port instance. Because that instance was still attached to its node, a late drag could create stale or ghost connection state.

Pure Virtual port compatibility was also direction-dependent, even though the canvas can check compatibility from either side of a connection.

## Fix

- Decode `\n`, `\r`, `\t`, and `\\` in String Split separators.
- Detach retired dynamic ports from their node.
- Reject connections involving detached ports.
- Make pure Virtual compatibility symmetric while preserving the restrictions for virtual arrays.
- Preserve literal escape sequences by allowing `\\n` when a literal `\n` delimiter is intended.

## Validation

- Added regression tests for escaped separators.
- Added coverage for stale dynamic-port references.
- Added an exact connection test for `Array Value at Index String → 3D Text Geometry.Text`.
- All 13 focused tests pass.
- Fabric Editor production build succeeds.

vade:
my only thought on this, post merge, is maybe the characterSet api is a cleaner UI for users?

jvcleave:
yeah - that is what I use in my app with buttons for SPLIT WORDS, SPLIT COMMAS, SPLIT LINES, or RAW

PR #299, opened 2026-07-24
Add configurable String Split modes

jvcleave:
Add a serialized Split Mode settings picker with Exact Separator, Words, Commas, and Lines behavior while preserving the existing Separator inlet and legacy default.

Cover Unicode whitespace and newlines, trimming, empty-value removal, and exact-separator compatibility.

image

related #297

### DIFF
--- a/Fabric/Nodes/Parameters/String/BaseFormatStringNode.swift
+++ b/Fabric/Nodes/Parameters/String/BaseFormatStringNode.swift
@@ -1,0 +1,207 @@
+//
+//  BaseFormatStringNode.swift
+//  Fabric
+//
+//  Shared base for the nodes driven by a `{name:spec}` format string — String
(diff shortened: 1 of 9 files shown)


tobyspark:
I’m keen on the textual separator (“exact”, here) being the default, to pair with _String Join_. This is what this does, so great, I’m just flagging it’s more than legacy to my taste.

Also: thanks for the PR!

tobyspark:
- rebuilt using strategy node
- UX pass, resolving the naming tension between “by x” and “as x”.
- alignment and correctness of string parsing and evaluation here, in _String Join_, _String Scanner_ and _String Formatter_

tobyspark:
Screenshot 2026-07-29 at 20 02 54

tobyspark:
Thanks @jvcleave !

Note existing documents won’t re-serialize correctly but migration is on the horizon.

PR #300, opened 2026-07-24
Feat – Control first execution

tobyspark:
Via Fable, having been tasked to look at architectural rationalisations after the added complexity Switch/Gate/Matrix introduced.

Removing the update-phase planning pass may not be compatible with current Satin prototype multi-threading work, but given this work has been done, posting here whether for study at minimum, or ultimately a merge.

> ## Summary
> 
> Removes the one-frame routing cold-start latency by resolving a routing node's control inputs (Index, map) before its route selection is read, and simplifies the GraphRenderer to a single inline execution path in the process. Both previously `.disabled` cold-start tests are re-enabled and pass: a Switch or Gate with a connected index routes correctly on the very first frame.
> 
> This reworks the execution core from the port-level dependency rework (`3ee85258`) — the update-phase planning pass is deleted entirely — which is why it's on its own branch rather than folded into the routing-node branch. Design rationale and the full tier history live in `FABRIC_EXECUTION_ARCHITECTURE.md` (project root, alongside the other audit docs).
> 
> ## Problem
> 
> Every routing pain point on the base branch traces to one circularity: **which edges are live this frame is a function of values computed during the same traversal that needs to know which edges are live.** A Gate's Index was pulled lazily inside the very pass asking the Gate which branch to pull, so:
> 
> - First frame after creation/load chose the wrong branch (index port still empty), self-correcting a frame later — the documented cold-start latency, with two `.disabled` tests asserting the behaviour we actually wanted.
> - The renderer needed keep-alive machinery (`PullResponse.declined(keepAlive:)`, the `.keepAliveWalked` state) purely so declined pulls couldn't starve control chains.
> - Routing correctness was traversal-order-sensitive by nature, guarded rather than guaranteed.
> 
> ## What changed
> 
> ### 1. `14778b03` — One inline execution path
> 
> The live view previously planned in `update()` (via a pullNode closure appending to `scheduledNodes`) and executed in `draw()`, while exporters/subgraphs/iterators executed inline during traversal. The split is *why* selections read stale values — the plan was built before anything executed — and the two closure parameters on `pullNode` existed only to serve it.
> 
> `draw()` now runs the same `executeAndDraw` traversal as every other caller: nodes execute depth-first as their dependencies resolve; pending resizes and `markClean` apply at the execution site. Deleted: `scheduledNodes`, `updateExecutionPlan`, `pendingSceneSync`, the `onTraverse`/`visit` closures, and the feedback cache's `cacheProcessedOutputs` split. Net −77 lines. Public API for exporters/subgraphs (`execute(graph:...)`, `executeAndDraw`) is unchanged.
> 
> Note: the originally proposed alternative — materialize the traversal into a `[Node]` plan and run it — turns out to be incompatible with fixing cold-start at all: building a plan queries route selections, and correct selections need values produced *this* pass. Inline execution isn't a compromise; it's the enabling move.
> 
> ### 2. `ca6028c9` — Control-first pulls (the actual fix)
> 
> `Node` gains `controlInputPorts` (default `[]`; `RoutingNode` → `[inputIndex]`, `MatrixSwitchNode` → `[inputMap]`). `pullNode` resolves these first — pulling and executing their upstream chains inline — *before* calling `respondToPull`. Route selection therefore always reads values produced this pass.
> 
> This is per-node resolution inside the recursion, not a global "control phase": no control-subgraph registry, laziness preserved (chains feeding routing nodes nothing consumes never run), and nested routing — a Switch inside another node's control chain — resolves naturally.
> 
> With controls resolved before the decline decision exists, the keep-alive machinery became redundant:
> 
> - `PullResponse.declined` loses its `keepAlive` payload.
> - `.keepAliveWalked` is deleted; a declined node returns to `.unprocessed`, so a later pull for a live output still evaluates it fully.
> - Feedback injection is split out of `setProcessingState` into an explicit `injectFeedback(forInlets:)`, called for control inlets and data inlets as each set becomes known.
> 
> ### 3. `a40d7e20` — Per-pass state without dictionaries
> 
> The per-pass costs on the render path: a `[UUID: NodeProcessingState]` reallocated from empty every pass (UUID hash per read/write), and port feedback values keyed by `(portID, frameNumber)` with a filter-allocating prune sweep per frame.
> 
> - Nodes claim a per-graph `executionSlot` when adopted by a graph — assigned in the `graph` reference's `didSet`, a single choke point covering `addNode`, document decode, and undo-restore. Slots are never reused.
> - The feedback cache keeps per-pass node state in a flat array indexed by slot, validated by a pass generation counter. `resetCacheFor` invalidates everything by bumping the generation: no allocation, no hashing.
> - Feedback candidates hold the upstream `Node` directly, making the back-edge check an array read.
> - Port values move to two buffers keyed by port ID alone, swapped on frame change.
> 
> ### 4. `b7482c13` — Published-flag invalidation (closes deferred review finding #7)
> 
> `graph.publishedOutputPorts()` caches against `connectionRevision`, and with the single inline path that cache decides what executes at all — published output ports are evaluation roots. Every existing mutation path bumped the revision through a side door (context menu via `rebuildPublishedParameterGroup`, auto-layout via `addNode`, port rebuilds via `addDynamicPort`/`removePort`), but a bare `port.published` toggle left the cached roots stale until an unrelated topology change. A `didSet` on `Port.published` now invalidates at the source, so no mutation site has to remember to. Regression test toggles the flag bare and asserts the port becomes a root on the next pass.
> 
> ### 5. `f889ba16` — Slotless-node fallback
> 
> A node pulled via its connections but never adopted by a graph has no execution slot, so its per-pass state silently failed to stick — it re-executed once per consumer pull, and a cycle through it would recurse without terminating (the UUID-dictionary implementation tolerated the case). A small generation-validated fallback dictionary — empty in practice, since every added or decoded node claims a slot — restores the tolerance. Regression test pulls a deliberately never-added provider from two consumers and asserts one execution per pass.
> 
> ### 6. `ee294b63` — Dirty-system TODO resolved as "dirty stays"
> 
> The feedback cache carried a TODO theorizing that node `isDirty`/`markClean` duplicated the renderer's per-pass processing-state tracking and could fold into it. The investigation refutes it — they are orthogonal axes: processing state is per-pass traversal dedup, `isDirty` is the cross-frame change signal whose writers the traversal cannot see (inlet/parameter writes, async IO events that mark dirty from callbacks, JavaScriptNode settings changes, undo-restore, Processor-mode SubgraphNode aggregating inner-graph state). Deriving execution from "an upstream ran this pass" would break zero-work steady state — Providers execute every frame, and it is send-side value dedup plus dirty marking that stops them cascading through clean chains; a port-level substitute collides with the `valueDidChange` edge-detection contract ~20 nodes consume inside `execute`.
> 
> No behavioural change: the TODO is replaced with this resolution, and the execute gate now reads `nodeExecutionMode` once with `isDirty` consulted last (both are computed on SubgraphNode — mode filters proxy ports, `isDirty` walks the inner graph). The investigation also surfaced one real bug, documented in place for a decision with vade rather than changed: **`Graph.needsExecution` seeds its `reduce` with `true` and is therefore constant-true**, so Processor-mode SubgraphNodes always execute — the literal `false`-seed fix would freeze subgraphs that self-animate behind static inlets (inner Providers execute without marking dirty), and the commented-out hardcoded `true` at the adjacent `isDirty` override suggests the current behaviour was parked deliberately.
> 
> ## Behaviour changes (intended)
> 
> | Behaviour | Before | After |
> | --- | --- | --- |
> | First frame with connected Index/map | Wrong branch, corrects next frame | Correct branch immediately (tests re-enabled) |
> | Gate recovery when only an unselected output is consumed | Frame 1 | Frame 0 |
> | Feedback injection after a dropped frame | Silenced (value had to be cached at exactly frame−1) | Injects last executed frame's values |
> | Live-view camera | `renderEncoder.draw` used previous frame's `firstCamera` | Uses this frame's (same as exporter path) |
> 
> Unchanged, deliberately: consumers of unselected Gate branches freeze; unrouted Matrix outputs freeze their consumers; a node sits out a pass only when *every* upstream pull declined (pull-order-independent); a Gate pulled only via unselected outputs keeps its data input lazy (regression-pinned).
> 
> ## Testing
> 
> `swift test` at every commit: zero issues outside the known pre-existing `MTLCommandBufferErrorDomain` GPU-fault flakes in "Graph Execution" / "Graph Export Renderer" (headless-Metal environment issue, identical set at base). Routing Nodes suite fully green, including the two re-enabled cold-start tests, the declined-pull-laziness regression, and the new published-flag and slotless-fallback regressions.
> 
> Incidental finding while writing the new tests: the GPU faults originate in the *Satin scene draw* (harness `drawScene: true` / `renderEncoder.draw`), not graph execution — the execute-only harness path is reliably green. Fixing the headless scene draw would unmask those two suites' logic coverage.
> 
> ## Review notes
> 
> - The decline flow in `pullNode` re-asks `respondToPull` for `.processed` nodes pulled on another port (Matrix recomputes its inverse map then). Cheap; a per-pass memo is possible if profiling ever cares.

### DIFF
--- a/Fabric/Graph/Graph.swift
+++ b/Fabric/Graph/Graph.swift
@@ -48,6 +48,14 @@ internal import AnyCodable
         nodes.filter { nodeViewModels[$0.id]?.isSelected == true }
     }
 
+    // NOTE (2026-07-24): the `true` seed makes this constant-true, so
+    // Processor-mode SubgraphNodes (which gate their execution on this via
+    // their isDirty override) always execute. Seeding with `false` would be
+    // the literal fix but a behaviour change: a subgraph whose inner graph
+    // self-animates (e.g. a time provider) while its outer inlets are static
+    // would freeze, since inner Providers execute without marking dirty. The
+    // commented-out hardcoded `true` at the isDirty override suggests this
+    // was parked deliberately — decide with vade before changing.
     var needsExecution:Bool {
         self.nodes.reduce(true) { (result, node) -> Bool in
             result || node.isDirty
--- a/Fabric/Graph/Graph.swift
+++ b/Fabric/Graph/Graph.swift
@@ -101,7 +101,18 @@ internal import AnyCodable
         case portConnectionMap
         case notes
     }
-    
+
+    /// Next per-graph execution slot handed to an adopted node (see
+    /// Node.executionSlot). Slots are never reused; a node claims one via the
+    /// didSet on its graph reference.
+    @ObservationIgnored private var nextExecutionSlot: Int = 0
+
+    internal func claimExecutionSlot() -> Int
+    {
+        defer { nextExecutionSlot += 1 }
+        return nextExecutionSlot
+    }
+
     public init(context:Context)
     {
         self.scene = Object(context: context)
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -156,11 +156,13 @@ public class GraphRenderer : ViewRenderer
     /// Single recursive pull behind every execution path: upstream nodes are
     /// pulled depth-first and executed inline as soon as their own dependencies
     /// are satisfied, so downstream nodes always read values produced this
-    /// pass. Returns false when this pull contributes nothing fresh downstream
-    /// — the node declined the requested output port (its keep-alive control
-    /// inputs are still pulled, at most once per pass, so it can select a
-    /// different route later), or sat out the pass because every one of its
-    /// own upstream pulls declined.
+    /// pass. A node's control inputs (Index, map) are resolved before its
+    /// route selection is queried, so routing never lags a frame behind its
+    /// control values and control chains can never starve — even when every
+    /// consumed output declines. Returns false when this pull contributes
+    /// nothing fresh downstream: the node declined the requested output port,
+    /// or sat out the pass because every one of its own upstream pulls
+    /// declined.
     @discardableResult
     private func pullNode(feedbackCache: GraphRendererFeedbackCache,
                           node: Node,
@@ -170,48 +172,62 @@ public class GraphRenderer : ViewRenderer
                           commandBuffer: MTLCommandBuffer,
                           clearFlags: Bool) -> Bool
     {
-        switch node.respondToPull(requestedOutputPort: requestedOutputPort) {
-        case .declined(let keepAlivePorts):
-            // Unselected route: nothing flows downstream from this pull, but
-            // the node's control inputs (Index, map) must keep updating or it
-            // could never select a different route — a Gate whose selected
-            // output has no consumer would starve its own Index chain, the
-            // class fixed for Matrix Switch in e776d909. The node names those
-            // inputs as keepAlive; walk them once per pass, marked by
-            // .keepAliveWalked, which also terminates control chains that
-            // cycle back into another unselected output of this node.
-            if feedbackCache.processingState(forNode: node) == .unprocessed {
-                feedbackCache.setProcessingState(.keepAliveWalked,
-                                                 forNode: node,
-                                                 executionInfo: executionInfo)
-                for inputPort in keepAlivePorts {
-                    pullUpstreamNodes(of: inputPort,
-                                      feedbackCache: feedbackCache,
-                                      executionInfo: executionInfo,
-                                      renderPassDescriptor: renderPassDescriptor,
-                                      commandBuffer: commandBuffer,
-                                      clearFlags: clearFlags)
-                }
+        switch feedbackCache.processingState(forNode: node) {
+        case .processing:
+            // Mid-evaluation upstream in this very pull: a feedback back-edge.
+            // The inlet reads the injected previous-frame value.
+            return true
+        case .declined:
+            return false
+        case .processed:
+            // Already ran this pass; only the per-port routing question
+            // remains, answered against controls resolved during evaluation.
+            if case .declined = node.respondToPull(requestedOutputPort: requestedOutputPort) {
+                return false
             }
+            return true
+        case .unprocessed:
+            break
+        }
+
+        feedbackCache.setProcessingState(.processing, forNode: node, executionInfo: executionInfo)
+
+        var attemptedPullCount = 0
+        var activePullCount = 0
+
+        // Resolve control inputs first: their chains execute inline here, so
+        // the values routing depends on land on the ports before the node
+        // selects its route below. This is also what keeps control chains
+        // alive when every consumed output declines — the starvation class
+        // fixed for Matrix Switch in e776d909.
+        let controlPorts = node.controlInputPorts
+
+        if !controlPorts.isEmpty {
+            feedbackCache.injectFeedback(forInlets: controlPorts, executionInfo: executionInfo)
+
+            for controlPort in controlPorts {
+                let pulled = pullUpstreamNodes(of: controlPort,
+                                               feedbackCache: feedbackCache,
+                                               executionInfo: executionInfo,
+                                               renderPassDescriptor: renderPassDescriptor,
+                                               commandBuffer: commandBuffer,
+                                               clearFlags: clearFlags)
+                attemptedPullCount += pulled.attempted
+                activePullCount += pulled.active
+            }
+        }
+
+        switch node.respondToPull(requestedOutputPort: requestedOutputPort) {
+        case .declined:
+            // Unselected route: nothing flows downstream from this pull. The
+            // controls above stay resolved; the node itself did not run, so it
+            // returns to .unprocessed and a later pull for a live output still
+            // evaluates it fully.
+            feedbackCache.setProcessingState(.unprocessed, forNode: node, executionInfo: executionInfo)
             return false
 
         case .evaluate(let pullingPorts):
-            switch feedbackCache.processingState(forNode: node) {
-            case .processed, .processing:
-                return true
-            case .declined:
-                return false
-            case .unprocessed, .keepAliveWalked:
-                break
-            }
-
-            feedbackCache.setProcessingState(.processing,
-                                             forNode: node,
-                                             activeInputPorts: pullingPorts,
-                                             executionInfo: executionInfo)
-
-            var attemptedPullCount = 0
-            var activePullCount = 0
+            feedbackCache.injectFeedback(forInlets: pullingPorts, executionInfo: executionInfo)
 
             for inputPort in pullingPorts {
                 let pulled = pullUpstreamNodes(of: inputPort,
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -254,7 +254,18 @@ public class GraphRenderer : ViewRenderer
                 node.resize(size: renderEncoder.size, scaleFactor: resizeScaleFactor)
             }
 
-            if node.isDirty || node.nodeExecutionMode == .Consumer || node.nodeExecutionMode == .Provider {
+            // Consumers and Providers always run; Processors run only when
+            // dirty — isDirty is the cross-frame change signal (inlet writes,
+            // parameter edits, async events like MIDI, subgraph inner state),
+            // which the traversal cannot derive: send-side value dedup is what
+            // stops always-executing Providers cascading re-execution through
+            // clean downstream chains. Mode is read once — it can be computed
+            // (SubgraphNode derives it from its proxy ports) — and the
+            // potentially costly isDirty (SubgraphNode walks its inner graph)
+            // is consulted last.
+            let executionMode = node.nodeExecutionMode
+
+            if executionMode == .Consumer || executionMode == .Provider || node.isDirty {
 #if DEBUG
                 commandBuffer.pushDebugGroup(node.name)
 #endif
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -38,10 +38,6 @@ public class GraphRenderer : ViewRenderer
     private var graphRequiresResize: Bool = false
     public private(set) var resizeScaleFactor: Float = 1.0
 
-    // Pre-sorted node list built in update(), consumed in draw()
-    private var scheduledNodes: [Node] = []
-    private var pendingSceneSync = false
-
     // One feedback cache per graph/subgraph UUID to handle different execution cadences
     private var feedbackCaches: [UUID: GraphRendererFeedbackCache] = [:]
 
@@ -103,8 +99,6 @@ public class GraphRenderer : ViewRenderer
         )
         currentExecutionInfo = GraphExecutionInfo(timing: timing, eventInfo: pendingEventInfo)
         pendingEventInfo = nil
-
-        updateExecutionPlan()
     }
 
     override public func cleanup() {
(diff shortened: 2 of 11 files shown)


tobyspark:
FWIW, here’s the architecture doc. In general, I feel a lot more confident about Fable on this codebase than I did Opus. (In a way that was different to the full-stack stuff I was doing at the same time, where Opus had crossed the threshold of “could ask it big things”)

# Fabric Graph Execution — Architectural Simplification

**Date:** 2026-07-24
**Branches:** Tier 1 on `feat-activeNodePorts-switch-node` (merged upstream via PR #191); Tiers 2–3 + follow-ups on `feat-control-first-execution`, rebased onto that merge and pushed to origin
**Context:** After landing the routing-node execution model (Switch / Gate / Matrix Switch) and fixing the 15 code-review findings, the execution core had accreted enough compensating machinery to warrant an architectural review. This doc records the diagnosis and a three-tier simplification, now fully implemented. Tiers 2–3 rework the core introduced by the port-level dependency rework (`3ee85258`), so they are held on their own branch for review with vade.

---

## The root cause

Every concern this work surfaced — the one-frame cold-start routing latency, the Gate/Matrix control-chain starvation deadlocks, the `.declined` processing state, the (since removed) re-pull-with-nil special case, the (since removed) `activeInputPortsDependOnValues` cache escape hatch — is one problem in five costumes:

> **The graph's effective topology (which edges are live this frame) is a function of values that are computed during the same traversal that needs the topology.**

A Gate's Index is pulled lazily inside the very pass that is asking the Gate which branch to pull. Everything else is compensating machinery for that circularity. This framing separates the cosmetic fixes from the changes that dissolve the machinery.

---

## Tier 1 — One pull API, per-inlet feedback caching ✅ implemented

Landed as two commits on `feat-activeNodePorts-switch-node`:

### `b96c6cce` — Cache feedback candidates per inlet, removing `activeInputPortsDependOnValues`

The feedback candidate cache keyed on (node, requested output port) and stored the node-level list of active-inlet pairings. For routing nodes that list follows runtime values (Index / map), so it could go stale without any topology change firing the invalidation. `activeInputPortsDependOnValues` existed solely as an opt-out flag — a trap API: any future node whose active-inlet set reads values but forgets the flag override silently gets stale feedback injection into a previously selected inlet.

What the cache actually saves per inlet — its first upstream outlet and that outlet's node — is **pure topology**. So it is now cached exactly that way: keyed by inlet ID, negative results included, invalidated on topology change; the active-inlet *set* is queried fresh on every use. The flag is deleted and the staleness class is unrepresentable rather than opted out of.

### `3c88424b` — Merge `shouldEvaluate` and `activeInputPorts` into a single `respondToPull`

The two methods were one concept answered across two virtual calls (Gate implemented one by calling the other; Switch re-encoded the same guard inline), and declining left the renderer to guess what a declined node still needs — `pullNode` re-entered itself with a nil requested port so the whole node evaluated just to keep the Index chain alive.

The contract became explicit and single-dispatch: `respondToPull(requestedOutputPort:) -> PullResponse`, with `.evaluate(pulling:)` and (at this tier) `.declined(keepAlive:)` — *a declining node names the inputs that must stay alive*. This deliberately sharpened one behaviour: a Gate pulled only via unselected outputs no longer evaluates its data-input branch or executes — nothing consumes the result. The regression test *"A declined pull keeps the control chain alive without evaluating the data input"* pins the laziness. (Tier 3 later simplifies this contract further — see below.)

---

## Tier 2 — One inline execution path; the closures cease to exist ✅ implemented

### `14778b03` — Unify the live draw path onto the single inline pull traversal (Tier 2a)

The renderer ran two execution models: `update()` built a schedule via `pullNode` with an append-to-plan closure and `draw()` executed it a phase later, while the on-demand path (exporters, subgraphs, iterators) traversed with an execute-immediately closure plus an `onTraverse` closure for pending resizes. The closures existed only to parameterize that split — and the split itself is why routing selections always read stale control values: the plan was built before anything executed.

The original proposal here was to *materialize the plan* (traversal produces `[Node]`, both paths run it). Implementation surfaced a cleaner resolution: a materialized plan is fundamentally incompatible with fixing the cold-start latency, because building a plan requires querying route selections, and correct selections require control values *produced this pass* — i.e. execution during traversal. So instead of plan-then-run, **everything became the inline path**: `draw()` now runs the same `executeAndDraw` traversal as every other caller, nodes execute depth-first as their dependencies resolve, and resize/markClean apply at the execution site. The closure parameters are deleted along with `scheduledNodes`, `updateExecutionPlan`, `pendingSceneSync` and the feedback cache's `cacheProcessedOutputs` split (net −77 lines). The closures were not converted to functions; the behavioural parameterization they served no longer exists.

(Plan *caching* — skipping traversal for static graphs — is thereby off the table, but the traversal itself is cheap once Tier 2b removes the dictionary churn; profile before mourning it.)

### `a40d7e20` — Slot-stamped node state and a double-buffered port cache (Tier 2b)

The per-pass costs that would actually show in Instruments: `nodeProcessingStateCache` was a `[UUID: State]` reallocated from empty every pass with a UUID hash per state read/write, and `previousFrameCache` keyed port values by `(portID, frameNumber)` with a filter-allocating prune sweep each frame.

Now: every node claims a per-graph `executionSlot` when adopted by a graph (assigned in the `graph` reference's `didSet` — one choke point covering addNode, document decode and undo-restore; slots never reused). The feedback cache keeps per-pass state in a flat array indexed by slot, validated by a pass **generation counter** — `resetCacheFor` invalidates everything by bumping the generation: no allocation, no hashing, O(1). Feedback candidates hold the upstream `Node` directly so the back-edge check is an array read rather than a second dictionary lookup. Port values live in two buffers keyed by port ID alone, swapped on frame change. One deliberate improvement: injection previously required a value cached at exactly frame-1, so a dropped frame silenced feedback for a frame; the previous buffer now carries the last executed frame's values across gaps.

---

## Tier 3 — Control-first pulls: resolve routing controls before route selection ✅ implemented

### `ca6028c9` — Resolve control inputs before route selection

The proposal sketched a *global* three-phase frame (control phase → plan phase → run phase). The implementation landed something both smaller and stronger: **per-node control-first resolution inside the single inline traversal**. Nodes declare `controlInputPorts` (RoutingNode: Index; Matrix Switch: map), and `pullNode` resolves them first — their upstream chains pull and execute inline — before `respondToPull` is asked. A route selection therefore always reads values produced this pass.

Why per-node beats the global phases: no control-subgraph identification pass, no early execution of control chains feeding routing nodes nothing consumes (laziness is preserved — an unreachable Gate's chain never runs), and nested routing (a Switch inside another node's control chain) resolves naturally through the recursion. The global sketch would have needed all three, plus a static control-root registry.

What it eliminated, as predicted:

- **The one-frame cold-start latency.** Both `.disabled` cold-start tests re-enabled and pass: first-frame routing is correct, and a Gate with a connected index recovers routing on frame 0 rather than frame 1.
- **All keep-alive machinery.** Controls are resolved before the decline decision exists, so `PullResponse.declined` lost its `keepAlive` payload, `.keepAliveWalked` is deleted, and a declined node simply returns to `.unprocessed` (a later pull for a live output evaluates it fully). Feedback injection became an explicit `injectFeedback(forInlets:)`, called for control inlets and data inlets as each set becomes known.
- **Pull-order dependence** for routing correctness, by construction.

Decline propagation is unchanged and still wanted: a node sits out the pass only when every upstream pull declined, and consumers of unselected branches freeze — that is routing *semantics*, not compensating machinery.

The semantic rule from the proposal stands, now by contract rather than traversal accident: a control input that depends on routed data of its own routing node reads previous-frame values via feedback injection.

---

## Follow-ups landed after the tiers

- **`b7482c13` — Published-flag invalidation (closes the deferred `publishedOutputPorts` review finding #7).** `graph.publishedOutputPorts()` caches against `connectionRevision`, and with the single inline path that cache decides what executes at all. Every existing mutation path bumped the revision through a side door (context menu via `rebuildPublishedParameterGroup`, auto-layout via `addNode`, port rebuilds via `addDynamicPort`/`removePort`), but a bare `port.published` toggle left the roots stale. A `didSet` on `Port.published` now invalidates at the source; regression test toggles the flag bare and asserts the port becomes a root next pass.
- **`f889ba16` — Slotless-node fallback.** A node pulled via connections but never adopted by a graph has no execution slot; without a home for its per-pass state it re-executed per consumer and a cycle through it wouldn't terminate (the UUID-dictionary implementation tolerated this). A small generation-validated fallback dictionary — empty in practice — restores the tolerance.

## Dirty-system collapse — investigated, resolved as "dirty stays" (`ee294b63`)

The proposal (and the cache's own TODO) theorized that `isDirty`/`markClean` duplicated the renderer's processing-state tracking and could fold into renderer-computed change propagation. The investigation refutes it — they are orthogonal axes:

- **Processing state is per-pass traversal dedup; `isDirty` is the cross-frame change signal.** Its writers are things the traversal cannot see: inlet/parameter writes (`NodePort.value` didSet), async events (MIDI/OSC/HID/GameController mark dirty from callbacks), JavaScriptNode settings changes, undo-restore, and Processor-mode `SubgraphNode` aggregating inner-graph state (its `nodeExecutionMode` is *dynamic* — Processor when it has published inlets and outlets).
- **"Any upstream executed → execute" breaks zero-work steady state.** Providers execute every frame; it is send-side value dedup plus dirty marking that stops them cascading re-execution through clean downstream chains. Renderer-derived propagation would run every graph downstream of any provider, every frame.
- **The port-level alternative rebuilds the same signal with a contract collision.** Gating on "changed since last execute" per port collides with `Port.valueDidChange`, which ~20 nodes consume *inside* `execute` for edge detection with their own clearing discipline.

What landed instead: the TODO replaced with this resolution; the execute gate reads `nodeExecutionMode` once and consults `isDirty` last (both can be computed — SubgraphNode filters proxy ports for its mode and walks its inner graph for `isDirty`); and one genuine bug documented in place for a decision with vade: **`Graph.needsExecution` seeds its reduce with `true` and is therefore constant-true**, so Processor-mode SubgraphNodes always execute — the literal fix (`false` seed) would freeze subgraphs that self-animate behind static inlets, and the commented-out hardcoded `true` at the `isDirty` override suggests the current behaviour was parked deliberately.

## Remaining opportunities (not taken)

- **`respondToPull` on processed re-pulls.** A `.processed` node re-pulled for another port answers `respondToPull` again (Matrix recomputes its inverse map). Cheap, but a per-pass memo could remove it if profiling ever cares.
- **Profiling.** The Tier 2b claims are structural (allocation and hashing removed); an Instruments pass on a realistic graph should confirm before further constant-factor work.
- **GPU-fault flakes: root cause located.** The headless `MTLCommandBufferErrorDomain` address faults come from the *Satin scene draw* (`GraphExecutionTestHarness` with `drawScene: true` / `renderEncoder.draw`), not from graph execution — the execute-only harness path is reliably green at any render size. Fixing the headless scene draw would unmask the "Graph Execution" / "Graph Export Renderer" suites' logic coverage.

---

## Sequencing / status

| Tier | Commits | Status |
|------|---------|--------|
| 1 — `respondToPull` + per-inlet feedback cache | `b96c6cce`, `3c88424b` on `feat-activeNodePorts-switch-node` | ✅ Merged upstream (PR #191) |
| 2 — Single inline path; slot-stamped state, double-buffered cache | `14778b03`, `a40d7e20` on `feat-control-first-execution` | ✅ Landed |
| 3 — Control-first pulls, cold-start latency removed | `ca6028c9` on `feat-control-first-execution` | ✅ Landed |
| Follow-ups — published-flag invalidation, slotless fallback | `b7482c13`, `f889ba16` on `feat-control-first-execution` | ✅ Landed |
| Dirty-system investigation — resolved as "dirty stays", gate tidied, `needsExecution` bug documented | `ee294b63` on `feat-control-first-execution` | ✅ Landed |

`feat-control-first-execution` reworks the `3ee85258` execution core (it deletes the update-phase planning pass entirely), so it lives on its own branch — rebased onto the PR #191 merge and pushed to origin — for review with vade before merging.

Test gate throughout: `swift test` in the Fabric repo; known pre-existing environmental flakes are the `MTLCommandBufferErrorDomain` GPU-fault family in "Graph Execution" / "Graph Export Renderer" — gate on no new failures outside that family. Everything above landed with zero non-GPU-fault issues; the Routing Nodes suite (both cold-start tests live again) is fully green.

tobyspark:
Superseded by #309

PR #301, opened 2026-07-25
Array Sequencer node

jvcleave:
Array Sequencer that loops an array over a configurable cycle duration using graph time or an optional external Time connection, with loop, reset, value, index, and progress controls.

This was my use case:

image

From what I could tell this would take quite a few other nodes to achieve. 

Cycle duration could use some clarification in my opinion (is this frames, seconds?) but will leave that open for debate

### DIFF
--- a/Fabric/Nodes/NodeRegistry.swift
+++ b/Fabric/Nodes/NodeRegistry.swift
@@ -451,6 +451,7 @@ public class NodeRegistry {
         ArrayFirstValueNode.self,
         ArrayLastValueNode.self,
         ArrayIndexValueNode.self,
+        ArraySequencerNode.self,
         ArrayAppendNode.self,
         ArrayReplaceValueAtIndexNode.self,
         ArraySplitAtIndexNode.self,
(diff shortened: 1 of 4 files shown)


tobyspark:
I think you needing to make this node is a failure on Fabric. It's missing some rather obvious nodes like LFO. The unserved need here is generating a number series, which isn’t unique to arrays. As well as something waveform-like, I have a signal based random node which hasn’t made it upstream yet, and something like a step sequencer would also be good. The one thing particular to arrays would be the ability to go through indexes randomly while not repeating one until exhausted. 

In sum, I’m going to take this PR as the prompt to make a PR with a more composable set of nodes around generating/sequencing numbers.

tobyspark:
...or have I missed some subtlety?

tobyspark:
Does #302 serve your use case? e.g. Pulse → Random Index → Value at Array Index

Note Random Index has a fair bit of configurability.

jvcleave:
with `Pulse → Random Index → Value at Array Index` I would still need to wire the array bounds correct?

maybe `Index Generator` is a better description for `Random Index` since it has non-random options?

I do like the option of being able to go through randomly without repeating.

tobyspark:
Yes, you'd need _Array Count_ in there too. 

Good one on _Index Generator_. _Random Index_ is where it started, and paired nicely with _Random Number_. But it has clearly outgrown that. Now I'm thinking what _Number Generator_ would have beyond random.

On not-repeating indexes... oh my the times I've implemented that ad-hoc in some interactive gubbins!

tobyspark:
Superseded by #302

PR #302, opened 2026-07-25
Feat – Node set: number and signal generators and improved smooth

tobyspark:
# Number generator & smoothing nodes

Adds a family of time-driven **Number** nodes for generating and shaping values, overhauls the existing `Smooth` node into a multi-mode, multi-direction filter, adds an optional **Clamp** to the existing `Number Remap` node, and introduces in-graph value visualization: a scope popover on `Trigger` and a new **Visualize** utility node.

## Summary

Six new nodes plus a reworked `Smooth`, registered in `NodeRegistry`:

| Node | Kind | Purpose |
|------|------|---------|
| **LFO** | Provider / TimeBase | Continuous oscillator, selectable waveform |
| **Pulse** | Provider / TimeBase | Periodic single-frame signal (Bool) |
| **Number Generator** | Processor | New value per signal — Random, Walk, or Even Spread |
| **Index Generator** | Processor | Index sequence per signal — random / shuffle / sequential |
| **Trigger** | Provider / TimeBase | Schmitt trigger with hysteresis + minimum on-duration |
| **Visualize** | Utility / Consumer | Scope plot of any numeric value, in Settings |
| **Smooth** *(overhaul)* | Provider / TimeBase | Multi-mode value smoothing, in any direction |

Terminology across the family: **signal** is the shared currency — Pulse emits one, the generators' `Signal` inputs consume one. Outlets are named for what they emit (LFO → `Number`, Trigger → `Trigger`). Strategy nodes title themselves with the active mode ("Random Number Generator", "One Euro Smooth").

## New nodes

**LFO** — outputs a unipolar `[0,1]` value cycling through a `Waveform` (Sine, Triangle, Sawtooth Up/Down, Square) once every `Period` seconds, with a `Phase` offset so several can run in sync but staggered. Phase is accumulated per frame from `deltaTime / period`, so changing `Period` live shifts the rate without jumping the output.

**Pulse** — emits a single-frame `true` (Bool) once every `Period` seconds — a periodic signal source for the edge-driven nodes — with a `Phase` offset. Firing is detected as an integer cycle-boundary crossing, so a `Period` shorter than the frame interval collapses to one pulse.

**Number Generator** — emits a new value in `[0,1]` on each signal. It's a `StrategyNode` whose **Mode** setting picks how the next value is drawn (the modes want different parameter ports, so Mode lives in Settings): **Random** — independent draws shaped by a `Distribution` (Uniform / Gaussian / Triangular), with a `Minimum Change` window that guarantees consecutive values differ by at least that amount; **Walk** — a bounded random walk that takes a step of up to `Step Size` from the current value and reflects off `0` and `1`, for organic wandering rather than jumps; **Even Spread** — a low-discrepancy sequence that fills the range evenly instead of clustering.

**Index Generator** — emits an integer index in `[0, Size)` (`NodePort<Int>`) on each signal. A `StrategyNode` like the Number Generator: **Mode** (in Settings) selects how the next index is drawn — **Random** (repeats allowed), **Random, no immediate repeat**, **Shuffle** (a full permutation before any repeat, with no repeat across the reshuffle seam), or **Sequential**. The finite modes (Shuffle / Sequential) carry a `Loop` port choosing restart-vs-hold when the sequence completes; the never-ending Random modes omit it.

**Trigger** — a Schmitt trigger with a minimum on-duration: the output rises to `1` when `Target` crosses above `Trigger Threshold` and falls back to `0` once `Target` is below `Release Threshold` and at least `Minimum Duration` seconds have elapsed. Hysteresis prevents chatter; Minimum Duration enforces a debounce floor.

**Visualize** — a Log sibling with the same any-value inlet: Log prints, Visualize plots. Settings shows a scope trace of the input whenever it unboxes to a scalar numeric (Float as-is, Int converted, Bool as 0/1 so pulses read as square waves), with y-axis scale and current-value readout; non-plottable types fall back to a hint. The goal is visualising any data type with performant Satin rendering — this numeric SwiftUI scope is the first pass.

## Trigger settings popover

Settings on `Trigger` plots a rolling `Target` history behind dashed threshold lines. One drag sets both thresholds — start point is Trigger, current point is Release, clamped to preserve hysteresis — with the y-axis locked for the gesture so the mapping can't shift under the cursor; a plain click quick-sets both at the click point. Each gesture is a single undo step. A fire indicator flashes on each rising edge.

## Smooth (overhaul)

`Smooth` becomes a `StrategyNode` whose **Filter Mode** setting picks the algorithm (the modes want different parameter ports, so this lives in Settings, not on a wired port):

- **One Euro** — adaptive low-pass; a `Responsiveness` (beta) port loosens it during fast motion to cut lag.
- **Exponential** — a one-pole filter with a single decay time constant.
- **Slew** — a constant-rate glide that arrives along a straight ramp; `Smoothing` maps to seconds per unit of travel.
- **Spring** — a damped second-order follower with momentum; a `Damping` port sweeps from bouncy (overshoot) through critically damped to overdamped. Integrated with semi-implicit Euler and adaptive substepping, so it stays stable when stiff or at low frame rates.

A **Direction** input restricts smoothing to rises, falls, or both: **Down** smooths falls while tracking rises instantly (a VU-meter peak-with-decay), **Up** is the reverse, **Up & Down** is symmetric. On the instant direction the output snaps to the input and the active filter's state — including any spring momentum — is reset to match, so smoothing re-engages from the peak without a jump.

`Smoothing` is a single normalised `[0,1]` amount for every mode, mapped geometrically for an expressive range. The Settings pane shows the mode picker with reactive per-mode usage guidance beneath it.

## Number Remap (clamp)

Adds an optional **Clamp** toggle to the existing `Number Remap`. With Clamp enabled, values outside the given ranges map to the range ends instead of extrapolating past them — inputs outside `[Input Min, Input Max]` land on the output min/max. Off by default, so existing graphs keep the previous extrapolating behaviour. The clamp is taken against the lower/upper of the two endpoints, so an inverted input range still clamps sensibly.

## Implementation notes

- Time-driven nodes use `.Provider` execution mode so they tick every frame (needed for the periodic/decay behaviours to advance while inputs are static); pure input-reactive nodes are `.Processor`.
- Visualization is push-based, with no shared mutable state across threads: nodes send one sample per frame through a Combine subject while their popover is observing (closed popover = zero work), and the view owns the history in a fixed-capacity, O(1)-append ring buffer on the main thread. Redraw is driven by sample arrival, not a polling clock — there is no `TimelineView`. Shared machinery in `Fabric/Views/Nodes/ScopeVisualizer.swift`.
- `StrategyNode`'s node-generated title is now the active strategy; `NodeTitleView` appends the type name. Vector Compose/Decompose opt back out where the token is redundant.
- `OneEuroFilter` and `LowPassFilter` gain a `reset(to:)` used by the Direction snap.
- Tests pin the Pulse → Number Generator signal contract: one new value per pulse period (including through the live update/draw path), and no evaluation when nothing pulls the generator.
- Builds clean via SwiftPM (Swift 6); no new dependencies.

## Notes for reviewers

- **Serialization change:** `Smooth`'s single `Frequency` port is replaced by the per-mode parameter set plus a serialised strategy. Documents saved with the previously shipped `Smooth Number` will not round-trip that parameter.
- The Direction gate composes only with instantaneous-state filters; windowed algorithms (median/moving-average) are intentionally out of scope here.

## Screenshots

Screenshot 2026-07-29 at 10 46 45

Screenshot 2026-07-29 at 10 47 16


tobyspark:
Inspired by @jvcleave’s comment https://github.com/Fabric-Project/Fabric/pull/301#issuecomment-5079160756 the  _Random Number/Index_ nodes are now _Number/Index Generator_, with number expanded to have a few different modes.

tobyspark:
Testing these nodes in Fabric Editor exposed an issue that doesn't happen in tests – the number generator never outputs. 

The below is verified, and there’s a failing test to confirm. This PR can be merged if the generator nodes are changed to `.Provider` instead of `.Processor` and the test skipped, but that doesn’t feel good. That said, I’m not sure what the alternative is in terms of deferred execution, without all nodes having separate update and draw methods.

> ## Live `update`/`draw` scheduling drops edge-driven `.Processor` nodes
> 
> ### Summary
> In `GraphRenderer`'s live render path, an edge-driven `.Processor` node fed by a
> one-frame signal (e.g. `Pulse → Number Generator`) fires at most once, then never
> again. The generator emits its initial value, the Pulse keeps flashing its
> `inputSignal`, but no new values are produced and the downstream chain freezes.
> 
> ### Cause
> The live path splits the frame in two:
> 
> - `update()` → `updateExecutionPlan()` builds the frame's schedule by walking the
>   graph and appending each node **if it is `isDirty`** (or a `Consumer`/`Provider`).
> - `draw()` then **executes** that pre-built schedule.
> 
> So `isDirty` is sampled **before any node executes**. A `.Processor` is only dirtied
> when its upstream *runs* — which happens later, in `draw()`:
> 
> 1. Frame N `draw()`: `Pulse` emits `true`, dirtying the generator — but frame N's
>    schedule was already built, so the generator isn't in it.
> 2. The dirty flag persists (un-scheduled nodes aren't `markClean`ed), so the
>    generator is scheduled for frame **N+1**.
> 3. Frame N+1 `draw()` runs upstream-first: `Pulse` executes and has already reverted
>    to `false`; the generator reads `false`, sees no rising edge, and never fires.
> 
> With the generator's output frozen, downstream `.Processor` nodes (Remap →
> Orientation Compose → …) are never re-dirtied and drop out of the schedule
> permanently. `.Provider` nodes are unaffected — they're scheduled unconditionally,
> so they run on the same frame the Pulse fires.
> 
> The on-demand path (`execute(graph:)` — subgraphs, export, tests) executes **inline
> during traversal**, so upstreams run before the downstream dirty check. No divergence.
> 
> ### Reproduction
> `NumberGeneratorPulseTests.executionParadigmsAgree` drives one `Pulse → Number
> Generator` graph through both paths with identical faked timing:
> 
> - `execute(graph:)` (inline): **3** new values over 3 pulse periods.
> - `update`/`draw` (deferred): **0** — never fires.
> 
> The test asserts the correct, equal behaviour, so it fails on the deferred path today
> by design.

tobyspark:
From discussion on Discord
- @vade has patched GraphRenderer to have a single execution path (f039eb143971ed51f6ee35d3c0cc0eb864f3ae84)
- This is now rebased on that
- Final test commit updated for unified execution path, i.e. `NumberGeneratorPulseTests.executionParadigmsAgree` is now `liveUpdateDrawPathFiresPerPeriod`

tobyspark:
I’m really happy where this has landed vs. the starting point of wanting the things I need for *spark stage in. I think it’s a great set of core nodes.

Late addition is the Visualizer node – super-useful for inspecting numeric change-over-time, and doesn’t get bogged down like writing to the console can. There probably is an architectural pass where we think about this kind of feature in the round, and a design pass too, but we get a lot of value here and now with it as-is.

PR #303, opened 2026-07-26
Feature/satin 2.0 plugin interface redux

vade:
This PR 

* Updates Satin submodule to a branch that adds throwable to setup, update and draw, and adds a delegate method to vend errors thrown behind Apple api boundaries that dont support surfacing errors (see CAMetalDisplayLink / CAMetalDrawable stuff)

* Updates Document loading GraphRenderer, and Node api's to support throwing new Fabric Errors, which have both fatal and recoverable variants. 

* Updates Fabric Editor to surface fatal errors to users, and catch recoverable without any real issue. The goal to eventually allow the UI to surface specific nodes throwing during execution in the UI (ie warning icons or what not)

* Updates the Fabric framework to support a QC Plugin like api surface  area with a new `FabricPlugin` api,  allowing:
   - `FabricPlugins` to host multiple nodes
   - `FabricPlugins` to be auto discovered via the `NodeRegistry`
   - `FabricPlugins` to have author metadata, versions etc
   - Bundles existing nodes into a core plugin : `FabricCoreNodesPlugin`
   
   
 Notes / Design decisions
 
 * `FabricCoreNodesPlugin` isnt a separate build target, mostly to ease integration, but it is loaded / treated as one internally by the NodeRegistry
 
 * `FabricPlugin` can be loaded from  `/Library/Application Support/Fabric/Plugins` and " ~/Library/Application Support/Fabric/Plugins`
 
 * Recoverable Errors thrown - im not convinced the current design is the best, we can review and keep this as a starting point

### DIFF
--- a/Fabric Editor/Export/GraphMovieExporter.swift
+++ b/Fabric Editor/Export/GraphMovieExporter.swift
@@ -176,7 +176,7 @@ final class GraphMovieExporter {
         let colorTexture = try self.makeIntermediateColorTexture()
         let depthTexture = try self.makeIntermediateDepthTexture()
 
-        renderer.start()
+        try renderer.start()
 
         let frameDuration = self.frameDuration()
         let sourceColorSpace = CGColorSpace(name: CGColorSpace.linearSRGB) ?? CGColorSpaceCreateDeviceRGB()
@@ -254,13 +254,13 @@ final class GraphMovieExporter {
                 await progressHandler?(frameIndex + 1, frameCount)
             }
         } catch {
-            renderer.finish()
+            try? renderer.finish()
             writerInput.markAsFinished()
             writer.cancelWriting()
             throw error
         }
 
-        renderer.finish()
+        try renderer.finish()
         writerInput.markAsFinished()
 
         if let completionError = try await self.finishWriting(writer) {
--- a/Fabric Editor/FabricDocument.swift
+++ b/Fabric Editor/FabricDocument.swift
@@ -65,13 +65,13 @@ class FabricDocument: FileDocument
         // Time source
         let currentTimeNode = CurrentTimeNode(context: self.context)
 
-        currentTimeNode.enableExecution(renderer: self.renderer)
-        currentTimeNode.startExecution(renderer: self.renderer)
+        try? currentTimeNode.enableExecution(renderer: self.renderer)
+        try? currentTimeNode.startExecution(renderer: self.renderer)
         
         // Math expression: secs * speed
         let mathNode = MathExpressionNode(context: self.context, expression: "secs * speed")
-        mathNode.enableExecution(renderer: self.renderer)
-        mathNode.startExecution(renderer: self.renderer)
+        try? mathNode.enableExecution(renderer: self.renderer)
+        try? mathNode.startExecution(renderer: self.renderer)
 
         // Publish the 'speed' port with a default of 10
         let speedPort = mathNode.findPort(named: "speed", as: ParameterPort<Float>.self)!
@@ -84,31 +84,31 @@ class FabricDocument: FileDocument
         // ports are dynamic (added by StrategyNode), hence findPort below
         // instead of typed accessor properties.
         let eulerNode = ComposeOrientationNode(context: self.context)
-        eulerNode.enableExecution(renderer: self.renderer)
-        eulerNode.startExecution(renderer: self.renderer)
+        try? eulerNode.enableExecution(renderer: self.renderer)
+        try? eulerNode.startExecution(renderer: self.renderer)
 
         // Geometry, material, mesh
         let boxNode = BoxGeometryNode(context: self.context)
-        boxNode.enableExecution(renderer: self.renderer)
-        boxNode.startExecution(renderer: self.renderer)
+        try? boxNode.enableExecution(renderer: self.renderer)
+        try? boxNode.startExecution(renderer: self.renderer)
 
         let materialNode = StandardMaterialNode(context: self.context)
-        materialNode.enableExecution(renderer: self.renderer)
-        materialNode.startExecution(renderer: self.renderer)
+        try? materialNode.enableExecution(renderer: self.renderer)
+        try? materialNode.startExecution(renderer: self.renderer)
 
         let meshNode = MeshNode(context: self.context)
-        meshNode.enableExecution(renderer: self.renderer)
-        meshNode.startExecution(renderer: self.renderer)
+        try? meshNode.enableExecution(renderer: self.renderer)
+        try? meshNode.startExecution(renderer: self.renderer)
 
         // Camera and light
         let cameraNode = PerspectiveCameraNode(context: self.context)
-        cameraNode.enableExecution(renderer: self.renderer)
-        cameraNode.startExecution(renderer: self.renderer)
+        try? cameraNode.enableExecution(renderer: self.renderer)
+        try? cameraNode.startExecution(renderer: self.renderer)
         cameraNode.inputPosition.value = simd_float3(0, 0, 3)
 
         let directionalLightNode = DirectionalLightNode(context: self.context)
-        directionalLightNode.enableExecution(renderer: self.renderer)
-        directionalLightNode.startExecution(renderer: self.renderer)
+        try? directionalLightNode.enableExecution(renderer: self.renderer)
+        try? directionalLightNode.startExecution(renderer: self.renderer)
         directionalLightNode.inputPosition.value = SIMD3<Float>(1, 2, 5)
 
         // Connections — animation chain. The Math Expression node derives its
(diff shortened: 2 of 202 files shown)


tobyspark:
This is great to see. I’ve tested this by trying to update Spark Stage to use it. 

The spec is
- Spark Stage Plugin bundle has to be built as part of the Spark Stage project
- Bundle has to be loadable by Fabric Editor (for fabric document authorship) and Spark Stage (for fabric document rendering)
- Spark Stage needs programmatic access to the nodes in Fabric core plugin and Spark Stage plugin.

Doing that hit a bunch of things. Headline is there are some structural things required within Fabric once you’re outside of Fabric, like node base classes being `open`, Fabric being linked dynamically. There’s a genie document below, but the headline is I think we need a public, Fabric-org-owned example app that embeds Fabric and has a custom plugin. And I think that in the sense of “proceeding without it is unviable” – the alternative is we lean on what I say about my Spark Stage experience, but much better to own this for the dogfooding and the example code for prospective adopters of Fabric.

We once had the idea of a _Fabric Cam_ app as an example of how to use the renderer. It was a good one. Let’s do that, and roll in a Plugin too.

tobyspark:
Hold your nose –

> ## 2. The gaps, in the order they surfaced
> 
> Each gap blocked the next. Several are now fixed on the Fabric branch; the rest are
> decisions or deferred work.
> 
> ### G1 — No public in-process plugin registration
> `PluginLoader` can load the embedded core plugin (`loadEmbeddedCorePlugin`, **private**,
> hard-wired to `FabricCoreNodesPlugin`) or external `.fabricplugin` bundles from disk. There
> is **no public API** for a statically-linked host to register a compiled-in plugin.
> - **Resolved by design choice, not code:** the shared-with-Editor requirement forces a real
>   loadable bundle anyway (see §3), so in-process registration isn't needed here. Still a
>   genuine gap for the simpler "app just wants to add a few nodes" case.
> 
> ### G2 — Missing nodes drop silently
> `Graph` decode printed "Failed to find nodeClass" and continued, dropping the node and its
> connections. A document authored against a not-installed plugin loaded as a quietly broken
> graph. **Fixed** (commit: throw on missing node type).
> 
> ### G3 — No record of which plugins a document needs
> A serialized graph stored only node *type names* (`String(describing:)`); nothing recorded
> which plugin a type came from, so a host couldn't verify up front that a document's plugins
> are present. **Fixed** (commit: persist a document's plugin requirements).
> 
> ### G4 — `SubgraphNode.subGraph` was `internal`
> A host (or a `SubgraphNode` subclass in another module) couldn't reach a subgraph's inner
> graph — needed e.g. to resolve a published proxy port back to the inner node it wraps.
> **Fixed** (commit: expose `SubgraphNode.subGraph` read-only).
> 
> ### G5 — The `Node` hierarchy was `public`, not `open` (the big one)
> The plugin system loads `Node` subclasses from bundles, but every node base class was
> `public class`. Swift forbids cross-module subclassing of non-`open` classes and
> cross-module overriding of non-`open` members — so **no code outside the Fabric module
> could define a node type at all.** The loader, `additionalNodeClasses()`, and
> `NSPrincipalClass` were plumbing with nothing able to author the thing they load.
> **Fixed** (commit: make the node authoring surface `open`). Also required making
> `BaseImageNode`'s `required init`s `public` (an `open` class needs accessible required
> inits).
> 
> ### G6 — Static linking duplicates Fabric → broken class identity
> A plugin bundle that statically links Fabric gets its **own copy** of every Fabric class.
> At runtime the host has another copy, so a node created by the bundle is **not** an instance
> of the host's `SubgraphNode` — `node as? SubgraphNode` returns nil, and the ObjC runtime
> warns about duplicate classes. Fatal for the loose-access design (§4).
> **Fixed** (commit: make Fabric a `.dynamic` library). The built bundle links
> `@rpath/Fabric.framework/…` and does **not** embed Fabric, so it binds to the host's single
> copy. Verified via `otool -L`.
> 
> ### G7 — ABI fragility across independently-built apps
> With a dynamic Fabric, a plugin loaded into a host binds to **the host's** Fabric, not the
> one it was built against. Swift has no stable ABI for ordinary frameworks (no
> `BUILD_LIBRARY_FOR_DISTRIBUTION` on Fabric), so a plugin built against Fabric vY loaded into
> an app shipping Fabric vX is only safe if vX≈vY. `FabricPluginAPIVersion` gates the *loader
> protocol*, not the *class ABI*.
> - **Decision (deferred):** don't pretend. Add a **build-stamp gate** — record the Fabric
>   version/commit the plugin was built against, and have the loader refuse-with-message on
>   mismatch rather than bind blindly. Library evolution (a stable plugin ABI) is the "real"
>   fix but is premature for a fast-moving Fabric.
> 
> ### G8 — Local-path dependency has no stampable identity
> The build-stamp gate only means something if "which Fabric" is an immutable, recorded value.
> A pinned **remote** package is recorded in `Package.resolved` with a revision hash; a
> **local path** (`../Fabric`) is resolved live from a mutable working tree and pinned to
> nothing. So the versioned-plugin story needs all consumers (app, bundle, Editor) to resolve
> Fabric from the **same pinned github commit**.
> - **Decision (deferred):** ship config is **pinned github + dynamic Fabric + build-stamp
>   gate**; local path is a **dev-loop-only** convenience. Provide a flip tool (a script over
>   the `xcodeproj` package reference, or an Xcode workspace local-package override, or a
>   SwiftPM mirror) so `use-local` / `use-pinned <commit>` swaps resolution *and* the stamp
>   mode (dev builds stamp `local-dev` → warn-only; pinned builds stamp the commit → enforce).
> 
> ---
> 
> ## 3. Why a real bundle, not compile-in
> 
> The app doesn't just host the nodes — it references them: `node as? StateSubgraphNode`,
> `[UUID: StateSubgraphNode]`, port proxies. That's a **compile-time** dependency, but a
> `.fabricplugin` is loaded at **runtime** and isn't linked by the app. And the same nodes must
> be authored in Fabric Editor, so compiling them into one app doesn't help the other. A shared
> framework or shared-source-in-two-targets both work but add ceremony and, for a framework,
> the duplicate-class hazard.
> 
> ---
> 
> ## 4. The key decision: loose access
> 
> Instead of giving the app compile-time plugin types, the host consumes plugin nodes **loosely**,
> by the surface Fabric already exposes and already uses elsewhere (`NodeRegistry.nodeClass(for:
> "SyphonClientNode")`, `availableNodes.first { $0.nodeName == "Fade Curve" }`):
> 
> - **Identity** by node **type name** (`"State Subgraph"`) or by "is a `SubgraphNode` carrying
>   the control ports", not by concrete Swift type.
> - **Ports** via `node.findPort(named:as:)` / iterating `node.ports` by `displayName`.
> - Hold nodes as `SubgraphNode` / `Node` (core types the app already links).
> 
> Consequences:
> - The plugin stays a **pure runtime `.fabricplugin` bundle** — no shared source, no shared
>   framework, no compile-time plugin types in the app, no in-process registration hook (G1).
> - The contract becomes **(node type name) + (port names)**: `"Enable"`, `"In Time"`,
>   `"Out Time"`, `"Active"`. Trade-off: a renamed port is a runtime miss, not a build error —
>   mitigate by centralizing those names as constants app-side.
> - Spark Stage's ~2 concrete touch-points (`discoverStageObjects`, `inGraphMoviePlayerID`) get
>   refactored to name/port access. `inGraphMoviePlayerID` needs the inner graph → G4.
> 
> The State nodes' subgraph behaviour maps cleanly onto the public surface: run children via the
> inherited `super.execute(...)` with an injected `userInfo` copy (carrying `StateToggleInfo`),
> and toggle visibility via the public `object` accessor — so the plugin never needs the
> `internal` innards, only what G4/G5 opened.
> 
> ---
> 
> ## 5. What is implemented (Fabric branch)
> 
> Five commits on `temp-plugin-number-generator-testing` (on top of the Satin-2.0 plugin merge),
> all building:
> 
> 1. **Persist a graph document's plugin requirements** — `GraphPluginRequirement`,
>    `PluginLoader.pluginRequirements(forNodeTypeNames:)`, `Graph.requiredPlugins` /
>    `missingPlugins`, encoded under a `pluginRequirements` key (`decodeIfPresent` = back-compat).
> 2. **Throw on missing node type** — decode throws `FabricError(.deserialization(.nodeNotFound))`,
>    naming the owning plugin via the manifest, instead of silent drop.
> 3. **Expose `SubgraphNode.subGraph`** read-only (`public private(set)`).
> 4. **Open the node authoring surface** — `Node`, `BaseObjectNode`/`ObjectNode`, `SubgraphNode`,
>    `BaseImageNode`, `BaseGeometryNode`, `BaseMaterialNode`, `BaseRenderableNode`, the blur
>    bases, `BaseTextureComputeProcessorNode`, and their documented override points → `open`;
>    `BaseImageNode` required inits → `public`.
> 5. **Fabric as a `.dynamic` library** — so host + plugin share one Fabric at runtime.

tobyspark:
The other thought I had about introducing plugins is that before we merge this we might want to think about nodes we have currently that perhaps should be separate plugins. I remember us discussing a while ago that maybe Syphon and HAP playback should be separate nodes, due to the framework dependency they bring.

vade:
Syphon yes, HAP no, i think video playback should be robust and out of the box just work imo.

PR #305, opened 2026-07-28
Feature – Plugin nodes can be renderable or subgraph types

tobyspark:
Latest changes to #303 solve most the structural issues flagged by the first round of *spark stage migrating to the new plugin API. The remaining gap is that the current plugin API does not allow plugins to provide renderable nodes (BaseObjectNode / BaseRenderableNode subclasses) or subgraph nodes (SubgraphNode subclasses).

This gap cannot be worked around on the plugin side: the graph discovers scene objects by `as? BaseObjectNode → getObject()` and subgraphs by `as? SubgraphNode`, so a node that doesn't subclass them is invisible to scene assembly and subgraph handling; and the base implementations call internal symbols (e.g. `invalidatePortCaches`, `portsChangedSubject`), so they can't be re-implemented against the public API either. The cause is that these base classes are public rather than open. This PR opens the object-node/renderable and subgraph node hierarchies (plus Node.teardown and Node.graph) for external subclassing — access widening only, no behavioural change. The alternative is to dispatch on a public protocol (e.g. SceneObjectProviding / SubgraphProviding) instead of concrete-class casts, letting plugins conform without subclassing — a larger change that inverts the current type-based dispatch.

### DIFF
--- a/Fabric/Graph/Node/Node.swift
+++ b/Fabric/Graph/Node/Node.swift
@@ -75,7 +75,7 @@ open class Node : Codable, Equatable, Identifiable, Hashable, Copyable
 
     public private(set) var context:Context
 
-    weak var graph:Graph?
+    public internal(set) weak var graph:Graph?
 
     // Method to register ports
     open class func registerPorts(context: Context) -> [(name: String, port: Port)] { [] }
--- a/Fabric/Nodes/Base/BaseObjectNode.swift
+++ b/Fabric/Nodes/Base/BaseObjectNode.swift
@@ -11,12 +11,12 @@ import Satin
 import simd
 import Metal
 
-public class BaseObjectNode : Node
+open class BaseObjectNode : Node
 {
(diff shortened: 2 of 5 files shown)


tobyspark:
Oh, [13dcb61](https://github.com/Fabric-Project/Fabric/pull/305/commits/13dcb611f963131fd7f13bb26ffee335bfc2ca11) is not a plugin limitation, but is similar in spirit – a longstanding spark stage reqs fix about opening Subgraph node so its usable by embedders. The use-case is not exotic: enables loading an existing fabric document as a subgraph into an API-created graph, and then can inspect that (sub-)graph.

vade:
I think this is good with one requested change - teardown remain internal, i believe stop, disable execution and deinit are good functions to use in lieu of exposing teardown.

tobyspark:
Reverted the `teardown` part, force-pushed. Verified my plugin can build, `deinit` is a better place anyway.

PR #307, opened 2026-07-29
Feature – Plugin loader resilience

tobyspark:
Further, potentially final, issues encountered and addressed in the process of making *spark stage adopt Fabric w/ plugins.

Two of the three commits are important and the title should be enough here
- [One plugin's failure should not cost the plugins behind it](https://github.com/Fabric-Project/Fabric/commit/79b207acf4717c2a7c58564989a7429b36435254)
- [Key the legacy node aliases on the spellings documents actually use](https://github.com/Fabric-Project/Fabric/commit/b1d2b56affc46906836b8bc185c162eeabfe6641)

### DIFF
--- a/Fabric/Nodes/NodeRegistry.swift
+++ b/Fabric/Nodes/NodeRegistry.swift
@@ -115,14 +115,21 @@ public class NodeRegistry
         registerCoreAlias("PassThroughNode<Satin.SatinGeometry>", PassThroughNode<Geometry>.self)
 
         // Structural array nodes were previously generic; old docs decode to type-agnostic versions.
+        //
+        // These suffixes must be spelled the way a *document* spells them, which is
+        // `String(describing:)` of the generic parameter — and that prints the
+        // underlying type, not the source alias. `simd_float2/3/4` are typealiases
+        // for `SIMD2/3/4<Float>`, so a saved graph reads `ArrayCountNode<SIMD3<Float>>`
+        // and never `ArrayCountNode<simd_float3>`. `simd_float4x4` is a struct in its
+        // own right, so it is spelled as written.
         let agnosticSuffixes = [
             "Bool",
             "Int",
             "Float",
             "String",
-            "simd_float2",
-            "simd_float3",
-            "simd_float4",
+            "SIMD2<Float>",
+            "SIMD3<Float>",
+            "SIMD4<Float>",
             "simd_float4x4",
             "FabricImage",
         ]
@@ -140,7 +147,7 @@ public class NodeRegistry
         }
 
         // Vector compose/decompose nodes were previously generic; old docs map to consolidated versions.
-        let vectorSuffixes = ["simd_float2", "simd_float3", "simd_float4"]
+        let vectorSuffixes = ["SIMD2<Float>", "SIMD3<Float>", "SIMD4<Float>"]
 
         for suffix in vectorSuffixes
         {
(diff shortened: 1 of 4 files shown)


tobyspark:
Looks like two remaining issues, both ok for dev but wanted for release. Copy and pasting this genie summary here as aide memoir for when I can pick this up. (Also: it’s late)

On the ‘Cannot consume Fabric as a remote SwiftPM package’, this would mean the repo holding the packages pinned to GitHub repo commit, and we have a local override we can turn if if we want it to pick up the local instead. 

> ---
> 1. Fabric can't be consumed as a remote SwiftPM package
> 
> Symptom. swift package resolve against any Fabric commit fails to resolve the Syphon and HapInAVFoundation binary targets.
> 
> Cause. Package.swift declares them by path:
> 
> .binaryTarget(name: "Syphon", path: "Frameworks/Syphon.xcframework")
> .binaryTarget(name: "HapInAVFoundation", path: "Frameworks/HapInAVFoundation.xcframework")
> 
> …but .gitignore:55 ignores Frameworks, and git ls-files Frameworks returns zero tracked files. The xcframeworks exist only as local build products of scripts/build-{syphon,hapinavfoundation}-xcframework.sh. A fetched checkout has nothing for those targets to point at.
> 
> Impact. Fabric can only be consumed as a local path with a pre-built Frameworks/. So there is no way to pin it: the app, a .fabricplugin, and Fabric Editor cannot be made to agree on one immutable Fabric revision. That also blocks the plugin build-stamp compatibility gate, which needs a stable "which Fabric was this built against" identity to compare.
> 
> Fix options. Commit the xcframeworks; or publish them as remote .binaryTarget(url:checksum:) — preferable, since it keeps binaries out of the history and gives them an integrity check.
> 
> ---
> 2. Plugin bundles embed binary-target frameworks they don't link
> 
> Symptom. A .fabricplugin ships copies of Syphon.framework and HapInAVFoundation.framework it never uses. Verified on SparkStageNodes.fabricplugin: both present in Contents/Frameworks/, while otool -L on the plugin binary matches neither.
> 
> Cause. A plugin must link the Fabric package product — that's what makes it bind to the host's single dynamic Fabric — and Xcode then embeds that package's binary targets into it. The embed is implicit, so there's no build-file entry to mark "Do Not Embed", and deleting the frameworks after the copy invalidates the bundle's code signature.
> 
> Impact. Dead weight in every Fabric plugin, and duplicated frameworks in any app that ships one, since the host already has its own copies. Not a correctness bug today — Fabric resolves Syphon from the host's copy via @rpath — so this is size and hygiene, not breakage.
> 
> Fix options. Largely toolchain behaviour, so the Fabric-side lever is to keep binary targets out of the product plugins have to link: e.g. split Syphon/Hap into a separate product or trait that hosts depend on and plugins don't. Worth confirming whether a supported per-target opt-out exists before restructuring.
> 
> ---

vade:
1 . `/scripts/` exists and the readme of the repository states you need to run them :) The genie should read readmes :) 

2.  There does appear to be a SwiftPM resource copy that is unnecessary - i noted this in plugin development, but it appears to be non breaking and just that, dead weight and requires some swift pm bullshit to resolve, and should be resolved but isnt a blocker to plugins at all. 

i'll review the code changes (first pass all seems sensible), but again, man, these LLM issues irk me - is there any way we can sanity check these - like, standard practice should always be "are you sure you arent full of shit, critique your own response because its 95% likely to be full of shit" :P

vade:
LGTM!

PR #309, opened 2026-07-30
Introduce Connections type - graph caches execution and connection topology.

vade:
This PR aims to solve 2 issues 

**Performance** 

Recent fixes to port dependency checking introduced additional execution overhead due to blindly evaluating graph dependencies each execution with no caching

**Modeling Correctness** 

With the recent intro of Matrix, Gate, Switch, we need to carefully and performantly model both connection topology, and execution topology

# Changes

`Connection` - which is a small helper class that captures both connection topology and execution topology managed by `Graph` during topology changes, and `Port`

`Node` / `Graph` gets helpers to manage  execution semantics for Gate/Switch/Matrix nodes via a new api `updateConnectionTopology` / `markConnectionTopologyChanged` respectedly.

`Graph` now caches connection, and calculates only when applicable, a node execution order that respects the execution / port dependency needs.

`GraphRender` now leverages `Graph`'s cached execution topology data order and does not need to calculate it on the fly every frame, but only when Graph has a change, dramatically increasing performance. 

This PR also 

* Adds code wise opt in **_Execution Tracing_** to `GraphRenderer`, which manufactures session of graph / node execution lists for GraphRenderer / Graph which can be serialized to disk, and used for testing, correctness, and performance understanding, and eventually perhaps GUI stepping and debugging in the editor. A Trace is started via `startExecution` and ends with `stopExecution` for now, and needs to be enabled in code, and on stop saves a JSON trace to `/tmp` - and is noted in the log.  This was used to trace correctness of implementation.  see 

* Fixes a bug in FeedbackCache that unintentionally used `Force` semantics.  

* Increases Texture Cache size and lifetime, avoiding expensive and slow cache trashing for many cases

* Updates GraphAutoLayout to uh, use a graph and respect connections, (tests updated too)

* Updates Satin which has some bug fixes

# Notes

All Satin tests pass, most Fabric tests pass save errant Timing based ones, which will be fixed external to this PR.

Im not 100% pleased with `Port` getting onValueChanged - it seems like a back door that can be abused, but is required for connection topology changes to be respected regardless of nodes being executed. 

Both `Port` and `Graph` having `[Connection]` arrays is a mild violation of single source of truth-iness, but i think for now its a good compromise.


Fabric/Graph/GraphRendererTextureCache.swift
```diff
@@ -39,7 +39,7 @@ internal final class GraphRendererTextureCache {
         public var framePoolCount: Int = 3
 
         /// When we need a new heap, allocate at least this many bytes (helps avoid tiny heaps).
-        public var minimumHeapSizeBytes: Int = 128 * 1024 * 1024 // 128 MB
+        public var minimumHeapSizeBytes: Int = 1024 * 1024 * 1024 // 128 MB
```
tobyspark: This is a big number. And maxHeapCount x16 makes it a _really_ big number.

vade: 1 gig of cache isnt that big on modern systems, but its fair, we can lower max heapCount to 4 or some such. The perf issue that motivated this change was lots of thrashing textures due to smaller heap sizes with low number of textures.

Im not convinced the heap design as is, is hear to stay. We can make some issues for better heap memory management design for improvement later. I'd be curious how MPS handles their temporary image implementation for example.

vade: One other quick note, i actually think max heap size is weird - if i have memory, i should use it. If ive got 64Gb of ram, i dont know that i want Fabric's runtime to scoff at allocating what it needs, vs thrashing memory and slowing things down. 

I imagine there's a better design pattern than a hard coded max heap count. I know it seems like its a lot, but the design should allocate only whats needed, when needed.

tobyspark: I think what really raised my eyebrow was that this value was _minimum_. That said, I don’t know what patterns or possibilities are for sizing caches up when there’s available memory but not otherwise.

tobyspark: ...wired memory, memory pressure, etc.

tobyspark:
I’m not precious about the above commits but they make sense / pass tests. The big thing that keeps on coming up is mutations on main while being read off-main in execute, a candidate or illustrative PR for that is #310

Fabric/Graph/GraphRendererTextureCache.swift
tobyspark: Final note here: wherever we land, lets make the inline comment agree.

vade:
wait, im confused - why are we worrying about threading in Fabric when we dont have threading solved in Satin?

tobyspark:
That’s one of many reasons why its a separate PR. Generally I’m keen on understanding threading issues between embedding app and a rendering thread.

PR #311, opened 2026-07-31
Feature – Port sync and declare then hydrate

### LINKED ISSUE #306: Documents recreate connections to nodes that no longer have that port
Save a document with a node with a connected port. Change the node interface to remove that port, rebuild, launch. Document shows the node still having the port, because the document load “reconnected” it. This is really only a dev issue, but it should not happen. 

Now we can throw, we should probably do that here (recoverable).

tobyspark:
Fixes #306. Graphs were fully serialised, bringing node-owned metadata (descriptions, ports, ...) along with genuine user-input. This keeps that, which could remain useful for context / compatibility. But previously on load, that would be blindly imported, regardless of any changes in the host Fabric. Old descriptions would not be updated, connections would be made to old ports the updated node has no way of knowing about. Now on load, nodes are created afresh and then hydrated with what’s relevant in the serialised graph.


vade:
So, a quick review, im fairly confident as is, many nodes wont properly decode, like `JavaScriptNode` and `LiveImageNode` which do not implement the new `self.updatePorts()` - but my gut is more will need it, like Math Expression, etc. I personally validated Live Image does not properly reload declared uniform ports for example (it does populate after opening and closing settings with no change though...)

So the above is just some mechanical work that I think was missed with this PR.

Some findings with help from Codex:

**Type-changed ports retain old UUIDs and can regain incompatible connections:**
- Node.swift (line 309)  matches only registry key and direction. Port.swift (line 213) then adopts the old UUID even when portType differs. Graph restoration finds that UUID, while NodePort.connect does not reliably reject incompatible types.

**ParameterPort identity is split after hydration.**
- A ParameterPort is initialized with port.id == parameter.id, but hydration mutates only Port.id. The backing parameter keeps its newly generated UUID. Even if nothing currently explodes, this violates the existing identity invariant and produces internally inconsistent encoded state. 
 

**Clamping during hydration destroys legitimate persisted state.**
- The implementation assumes every saved slider value originated in the slider UI. A slider-backed inlet may have been driven outside its UI range by a connection or procedural API. Connections have not been restored when clamping occurs, so the code cannot distinguish those cases. If min/max is a parameter invariant, enforce it in Satin/model assignment consistently; if it is UI presentation, hydration must not clamp it. The present compromise is the wrong layer.

That last one is admittedly arguable, but i'd say, if a Satin Parameter can encode / decode something out of range, and we agree thats bad (ie, no overdrive or whatever), then the fix should be in Satin Parameter IMO. (and FWIW i kind of like the ability to overdrive a slider but i get its perhaps unexpected).

Structurally, some thoughts from me:

* `PortRegistry` already owns UUID, ordering, registration, and snapshot reconciliation and i think this sort of violates separation of concerns or at min duplicates some effort that should be unified appropriately somewhere, if not PortRegistry.

* `Port` should be responsible for its hydration if possible, and `ParameterPort` should keep its parameter id in sync ideally

* `.droppedPortStateDiagnostics` isnt actually used anywhere from what i can tell, not in tests either? - id much rather just throw a specific resumable error that connections could not be restored or some such? 

My feeling here is a lot of stuff is just 'in the wrong place' from a separation of concerns perspective.

tobyspark:
Don’t review this yet.

tobyspark:
Ready to review. Some comments are more verbose than I’d like, but otherwise I’m happy with it. Does it pass your smell test @vade ? Are things where they should be?

vade:
Awesome, ill give it a look over when i can catch my breath! much obliged!

Issue #271, opened 2026-06-16
New Document: Use Number Integral, PassThroughNode to gain min/max for float type


tobyspark:
Not a priority, but easy wins –

Not using `Number Integral` is a whoopsie by me we shouldn’t be showing to any new users.

Having the speed published input be a slider would be a much better UX than the text field. To do so we should expose the min/max values already on the Satin/Fabric float parameter type in PassThroughNode e.g. 7bdea748910208a837a411315ed6f5250d28337b

PR #312, opened 2026-07-31
Feature – New document without the faux pas

tobyspark:
Putting our best foot forward for the new release...

Screenshot 2026-07-31 at 03 40 20


tobyspark:
...or at least, not having a faux-pas in there.

tobyspark:
Addresses half of #271.

PR #318, opened 2026-08-09
Nodes have a regular, canonical name for logging

tobyspark:
Prompted by hard-to-parse logs where e.g. the math expression was logged as the name. 

Two new fields on `Node` –
- `typeName` used in logs etc. where performance + brevity important. 
- `debugName` used in logs etc. elsewhere.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,15 +77,17 @@ For all development:
 
 ### 2.1  Nodes & Execution
 - Nodes define immutable static metadata:  `nodeType`, `nodeExecutionMode`, `nodeTimeMode`,  `name`, `nodeDescription`.
-- A node is named from three sources, two of them overridden by the subclass:
-  - `class var name` — override: the name the type is registered and listed under; `canonicalName` on an instance.
-  - `func deriveSubtitle() -> String?` — override where the node describes itself, nil otherwise: Math Expression's expression, StrategyNode's strategy. Read it off an instance as `subtitle`, final and normalized.
-  - `var userName: String?` — final: the user's rename, serialized.
-  - An empty name is no name: `subtitle` reads a `deriveSubtitle()` of "" as nil, and `userName` normalizes "" to nil at set and decode. Overrides and consumers never guard emptiness themselves.
+- A node is named from four sources, three of them overridden by the subclass:
+  - `class var name` — override: the stable name the type is registered and listed under, and the default instance title.
+  - `func deriveTitle() -> String` — override only when an instance has a more specific authoritative title, such as a shader-backed `BaseImageNode`; defaults to `Self.name`.
+  - `func deriveSubtitle() -> String?` — override where the node describes itself, nil otherwise: Math Expression's expression, StrategyNode's strategy.
+  - `var userName: String?` — final: the user's rename, serialized and public for rename-specific callers.
+  - An empty name is no name: `userName` normalizes "" to nil at set and decode.
 - From those it composes, both final:
-  - `var title` — what to call the node: `userName ?? canonicalName`. Note `subtitle` is deliberately absent: it accompanies the title, it does not replace it.
-  - `var debugDescription` — every name the node answers to, e.g. `Math Expression (sin(x) · My Rename)`. `print(node)` and `"\(node)"` give it: log a node directly rather than reaching for a name. Use `canonicalName` where brevity or per-frame cost matters.
-- Fire `subtitleSubject.send()` whenever state feeding `subtitle` changes; `NodeViewModel` mirrors the node's `title` / `canonicalName` / `subtitle` observably, and adds `titleLabel` (`userName ?? subtitle`) — the label all title UI draws ahead of the canonical name.
+  - `var title` — the normalized result of `deriveTitle()`, falling back to `Self.name` when empty.
+  - `var subtitle` — `userName ?? deriveSubtitle()`, normalized so empty values and values equal to `title` read as nil. General consumers read this rather than `userName`.
+  - `var debugDescription` — the title plus resolved subtitle, e.g. `Math Expression (My Rename)`. `print(node)` and `"\(node)"` give it; diagnostic labels and traces use it when they need both values.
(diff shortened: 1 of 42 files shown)


tobyspark:
@vade I didn’t touch the Graph execution trace. You may want to update the fields cached there as appropriate.

...this also rather begs the question what is `name` even for, now we have a custom display of the name in the Node’s UI. Further, the pattern that has emerged there is that `displayName` is better thought of as a refinement to the node’s type (class) name. So really the name heuristic should be ‘userName OR typeName (refinement) OR typeName’

tobyspark:
Prompts from Discord chat
- Potentially better API design if customName was a function, i.e. more obvious for a node to return something about its current state.
- Audit for adoption across all nodes etc.

vade:
Thanks for the updates!

I agree there are three useful concepts expressed here:

* The registered node name.
* An optional user rename.
* Optional node-derived context, such as an expression, strategy, or device.

I think these can be expressed more clearly as:

* class var name: the stable registered/type name.
* Instance name: userName ?? Self.name.
* subtitle: optional node-derived context for the UI.

The UI can render name and subtitle however it wants. Logging call sites can explicitly use the stable type name, instance name, or debugDescription, depending on their needs.

Thinking more deeply about the naming here, i think `customName` is ambiguous when next to `userName`, and `canonicalName` is especially confusing because it is mutable while the registry name is the actual canonical identity. Since `canonicalName` is only used in  diagnostic composition, I would put that formatting directly in `debugDescription` and lose `canonicalName ` completely.

I spent some time iron-manning your PR with codex, and it tended to agree canonicalName should likely go, or at a minimum not be called canonical (my original point, which i keep trying to make, is that the ONLY canonical name is the registryName, as it wont change unless the node class implementation changes, and thus is authoritative) 

Some code nit picks as is:

* `deriveCustomName()` is redundant to the property customName, as the property doesnt cache, and calls the function each time. The only thing it does is check for isEmpty, which could be rolled down stream and simplify the node api by dropping a field. 

* drop `canonicalName` completely and put that logic into `debugDescription` or if we have to have some function or property , call it something other than `canonicalName` because, it is mutating, which isnt canonical and could be mis-interpreted (I certainly would expect it to be authoritative)  

If we don't want to re-litigate the naming ( which, i'll be honest, the more i sit with it the less i like it as is), i think at a min we can factor / simplify this more.

tobyspark:
I also don’t like `customName`. It was the least worst option I could come up with – anything else I came up with seemed over-specialised or over-wrought (`typeName` + `typeNameRefinement`). I had thought of `subtitle` but then it seemed worse to have `subtitle` but no `title`. And `subname` seems weird, but maybe?

I do think the DX / API design is better with an instance field to get the class name. In short, the thing you want in most cases will auto-complete. I think it also names the intent better. Having to do `type(of: ).name` isn’t clearer to my mind, in a code base where I’ve also seen `String(describing: type(of:))` and/or `String(describing: )`.

So, moving the verbose instance name code into `debugDescription`, we could have
- `Self.name`
- `self.canonicalName = Self.name`
- `self.name = userName ?? canonicalName`
- `self.subname`

...subname vs. subtitle, dunno.

In terms of simplification – I've have a slight dread about this whole PR for a while (doubly so with the recent work beyond it to adopt `customName` across other nodes) as it feels like a simple logging win has metastasised. But I am reasonably confident that what I landed on is the most DRY _if we want a provably correct system_. The context these names sit in is complex – they come in from the UI, in from decode, in from node state changing. There is a view model in the mix. There is an execution model separate to UI update needs. With that further adoption work it’s less a question of taste, as naive implementations fail in the UI.

Coming back to the specifics in this and only this PR, I don’t think I’ve missed some kind of trick in missing simplicity. If we don’t guarantee optional names are `nil` at source, then there’s a whole lot of defensive work downstream. The move available to us is to accept not being watertight. We could say we prefer to accept that empty strings could confuse things, because we don’t think empty strings will enter the system, and if they do it’s because of e.g. legacy documents we don’t care about. If you want to take the current state of this PR and do a simplification pass I’d be happy to see it (as said above, I have no love this work any more!), but I don’t think you’d end up with something meaningfully different without trading on correctness or DRY.

(There’s a tiny bit of solace in strings being null vs. empty is a programming trope. Perhaps there’s a simplification move where the strings are not optional and so there is only `isEmpty`, but I prefer the null check)

tobyspark:
From [discussion on Discord](https://discordapp.com/channels/1443372127427366933/1505751776366690445/1538935520686510230), naming updated.

vade:
some style nits / convention nits - this still reads like weirdly put together LLM code if im being honest. I iron-manned my review notes after manual code review with Codex to arrive at the following notes:

- `userName` is used almost exclusively outside of the node internal API and is the de-facto api contract, but i think `subtitle` should be the public entry point, and not `userName`

Our naming flow is weird and not at all legible to 3rd parties IMO: 

- Self.name → `canonicalName`
-` userName` ?? `canonicalName` → `title`
- `deriveSubtitle()` → `subtitle`
- `userName` ?? `subtitle` → `titleLabel`
- In particular, the `title` UI does not use `Node.title`; it independently reconstructs another resolved label in NodeViewModel.

Could we instead define:

- `title` as the stable registered name: `Self.name`
- `subtitle` as `userName ?? deriveSubtitle()` with the isEmpty logic etc etc. 
- `userName` as final internal writable/serialized rename state, used only by rename-specific code

That would let us remove `canonicalName` entirely because it is only an alias for `Self.name`, and likely remove `titleLabel` because `subtitle` becomes the public resolved display value.

`debugDescription` can compose `title` and `subtitle` directly.

Logging sites that intentionally need only the stable registered identity can use `title`; rename-aware diagnostics can use `debugDescription`.

Overall this is moving in a better direction, but the API still feels like it exposes several intermediate values when callers only need a stable title and an optional resolved subtitle.

I know, im sorry im being a fucking stickler on this...

Also im not clear if you meant to push changes to your private branch or the fabric branch - there seems to be more stuff not in the PR, like postInit etc, which i have more thoughts on!

Issue #320, opened 2026-08-10
No straightforward way to set the background colour; QC had the _Clear_ patch


tobyspark:
> toby*spark — 20:20
> What's the simplest way of setting the background colour? In QC, there was the Clear patch.
> ...this would now be a property of a custom camera?
> vade — 20:23
> its part of the graph renderer which concocts the command buffer in metal. Last i looked at it, its sort of akward - i didnt find a  a clean way to do it from a patch - just not wired up. right now clear it manually set on a graph renderer.
> i think though, with the newer node api it might be doable, i havent revisisted it in a while 
> (I havent thought about that in a while)
> toby*spark — 20:23
> Gotcha.
> vade — 20:24
> also, clear isnt a 'render command’
> its part of render metadata to a render encoder for drawing -  so clear as semantically equivalent in QC, i dont think will happen
> it sort of means 'for the next draw what should i clear to' :X its a metal command buffer submission thing vs a state / command in old opengl
> i suppose you could encode an empty draw command, but its not quite the same
> (there isnt one canonical clear color any more either, its part of pass description shit)

https://discordapp.com/channels/1443372127427366933/1443372129083982069/1536454224483983471

PR #323, opened 2026-08-15
Fix/clear node

vade:
This PR adds a QC like `Clear` node to Fabric. 

Given Fabric isnt immediate mode rendering in OpenGL, there are some differences with this node

a) Clear node is setting the clear color within the render pass encoders next encode pass
b) clear does not have a render order, and does not actually clear anything
c) multiple clear patches -> the last one to execute wins. 

Some possible improvements should we want

a) it might make sense to call this 'background color' vs clear? I think i am leaning to that as it also disambiguates it doing work it isnt (ie actually clearing buffers).
b) if we do need or want a full screen overlay / clear with a color (or even blending), or full screen draw a color function, we could make a custom pass object - but it would need to respect Satin's scene graph execution ordering. 
c) Toby suggested making this a property of the camera, which, I dunno, strikes me as odd but i also dont claim to have all the answers. Maybe users can tell us?

At a minimum, this does do what it needs to do, and we can iterate on it in the future? 

This PR concretely:

* Adds a Clear Node
* Adds it to the standard plugin registry
* Updates Graph Renderer to cache the initialized clear value so on Clear Node deletion, we revert to the standard GraphRenderer's clear color.

### DIFF
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -165,6 +165,17 @@ public class GraphRenderer : ViewRenderer
 
     public func executeAndDraw(graph: Graph, executionInfo: GraphExecutionInfo, renderPassDescriptor: MTLRenderPassDescriptor, commandBuffer: MTLCommandBuffer) throws
     {
+        let clearColor = self.renderEncoder.clearColor
+        let colorLoadAction = self.renderEncoder.colorLoadAction
+        let depthLoadAction = self.renderEncoder.depthLoadAction
+        let stencilLoadAction = self.renderEncoder.stencilLoadAction
+        defer {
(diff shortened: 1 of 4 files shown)


vade:
Would resolve #320

tobyspark:
I think _Background Color_ is better than _Clear_, as it’s more searchable and doesn’t mislead about what it is.

On the broader design question, I am uneasy with the pattern of dropping “last one wins” context/environment nodes into the graph. QC’s _Clear_ patch made sense here with it being a discrete rendering step in the render order; this is not that though. My instinct is if there’s one and only one of something needed to render, i.e. the background colour here, we should be exposing that fact. And how that fits with the pull-from-consumer graph model (or not) is another discussion on that. There’s more here, though, ref the _Camera_ design discussion in #181 .

PR #324, opened 2026-08-17
Feature – Per node error isolation

tobyspark:
A single node throwing aborts further node execution for that frame. If it’s recoverable, node execution should continue. This PR does that.


tobyspark:
Note this is particularly important for nodes with edge detection that would have their execution skipped but were nonetheless marked clean

vade:
LGTM!

PR #325, opened 2026-08-18
Fix – An unwired ParameterPort falls back to its parameter

tobyspark:
As per title.

### DIFF
--- a/Fabric/Graph/Port/ParameterPort.swift
+++ b/Fabric/Graph/Port/ParameterPort.swift
@@ -119,6 +119,20 @@ public class ParameterPort<ParamValue : PortValueRepresentable & Codable & Hasha
     {
         didSet
         {
+            // A parameter-backed inlet has no valueless state: the parameter is
+            // its resting value. Disconnecting an upstream port force-sends nil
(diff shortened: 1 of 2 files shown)


Fabric/Graph/Port/ParameterPort.swift
```diff
@@ -119,6 +119,20 @@ public class ParameterPort<ParamValue : PortValueRepresentable & Codable & Hasha
...
+            // downstream so the inlet doesn't keep stale data, which would
+            // otherwise leave this port — and everything reading it — with a
+            // missing value for as long as it stays unwired. Re-seat on the
+            // parameter instead; the re-entrant assignment runs the change
+            // bookkeeping below.
+            guard let value = self.value
+            else
+            {
+                self.value = self._parameter.value
+                return
```
vade, line 133: should this path mark dirty or call onvaluechanged?

tobyspark: > The re-entrant assignment runs the change bookkeeping already there, so valueDidChange and markDirty still fire.
– commit message. 

Though this didn’t get a test, and because other things have been fixed I don’t have the symptoms any more. I’ll add a test.

PR #326, opened 2026-08-18
Fix – Make current camera system workable

tobyspark:
The current camera system is somewhat broken. Before attempting any feature additions or redesign – see #181 – this PR contains many small fixes that together make the current status quo at least workable.

### DIFF
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -211,9 +211,6 @@ public class GraphRenderer : ViewRenderer
             }
         }
 
-        // The graph's camera, or the free one — never the last camera seen. A
-        // graph whose only camera node is deleted has no camera again, and
-        // holding the deleted one is how it kept rendering through it.
         self.currentCamera = graph.latestCamera ?? self.defaultCamera
 
         var capturedError: (any Error)?
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -73,9 +73,9 @@ public class GraphRenderer : ViewRenderer
         self.renderEncoder.sortObjects = true
 
         self.sceneProxy = Object(context: context)
-        self.defaultCamera = PerspectiveCamera(context: context)
-        self.defaultCamera.position = simd_float3(0, 0, 2)
-        self.defaultCamera.lookAt(target: .zero)
+        // The camera a graph gets for free is the camera node's, at its
+        // defaults, so adding that node to a graph reframes nothing.
+        self.defaultCamera = PerspectiveCameraNode.makeDefaultCamera(context: context)
 
         self.privateTextureCache = GraphRendererTextureCache(device: context.device)
 
@@ -138,8 +138,10 @@ public class GraphRenderer : ViewRenderer
         self.renderEncoder.resize(size)
         self.graphRequiresResize = true
         self.resizeScaleFactor = scaleFactor
+        // Field of view is the camera's, not the drawable's: a resize changes
+        // the resolution rendered and how much is visible to the sides, and
+        // must not change the framing of what a graph was authored against.
         self.defaultCamera.aspect = size.width / size.height
-        self.defaultCamera.fov = radToDeg( 2.0 * atan( (size.height / size.width) / 2.0 ) )
     }
 
     // MARK: - Draw
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -211,7 +211,7 @@ public class GraphRenderer : ViewRenderer
             }
         }
 
-        self.currentCamera = graph.firstCamera ?? self.currentCamera ?? self.defaultCamera
+        self.currentCamera = graph.latestCamera ?? self.currentCamera ?? self.defaultCamera
 
         var capturedError: (any Error)?
         var scheduledNodes = nodesInExecutionOrder(for: graph)
--- a/Fabric/Graph/GraphRenderer.swift
+++ b/Fabric/Graph/GraphRenderer.swift
@@ -211,7 +211,10 @@ public class GraphRenderer : ViewRenderer
             }
         }
 
-        self.currentCamera = graph.latestCamera ?? self.currentCamera ?? self.defaultCamera
+        // The graph's camera, or the free one — never the last camera seen. A
+        // graph whose only camera node is deleted has no camera again, and
+        // holding the deleted one is how it kept rendering through it.
+        self.currentCamera = graph.latestCamera ?? self.defaultCamera
 
         var capturedError: (any Error)?
         var scheduledNodes = nodesInExecutionOrder(for: graph)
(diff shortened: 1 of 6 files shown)


Fabric/Graph/GraphRenderer.swift
```diff
@@ -199,7 +211,10 @@ public class GraphRenderer : ViewRenderer
             }
         }
 
-        self.currentCamera = graph.firstCamera ?? self.currentCamera ?? self.defaultCamera
+        // The graph's camera, or the free one — never the last camera seen. A
```
vade: doesnt this contradict the code in Graph.swift:654 / 660 ?

Fabric/Nodes/Object/Camera/OrthographicCameraNode.swift
```diff
@@ -16,20 +16,34 @@ public class OrthographicCameraNode : ObjectNode<OrthographicCamera>
...
-        let ports = super.registerPorts(context: context)
-        
+        // Position, scale and orientation carry the camera's own defaults and
+        // wording: an object's position default would put the camera on its
+        // subject, and its scale and orientation describe an object in a
+        // scene rather than the view of one.
+        let ports = super.registerPorts(context: context).filter { !["inputPosition", "inputScale", "inputOrientation"].contains($0.name) }
+
         return  [
-                    ("inputLookAt", ParameterPort(parameter:Float3Parameter("Look At", simd_float3(repeating:0), .inputfield, "Target position the camera points toward")) ),
```
vade, line 26: Why are we removing look at? Its super useful for making automatic camera moves and look at subjects or points of interest.

Fabric/Nodes/Object/Camera/OrthographicCameraNode.swift
```diff
@@ -16,20 +16,34 @@ public class OrthographicCameraNode : ObjectNode<OrthographicCamera>
...
+                    ("inputOrientation", ParameterPort(parameter:Float4Parameter("Orientation", defaultOrientation, .inputfield, "Which way the camera faces, as a quaternion (X, Y, Z, W). Identity looks along -Z")) ),
+                    ("inputScale", ParameterPort(parameter:Float3Parameter("Scale", simd_float3(repeating:1), .inputfield, "Scales the view volume, so the scene draws smaller as this grows: two world units tall at 1")) ),
```
vade, line 42: Why do we need this? Scale on camera strikes me as odd. If you want to scale objects in the world, why not use a transform on the objects. Camera scale has a smell to me and does not track any physical camera property nor do i believe its an intuitive control.

Fabric/Nodes/Object/Camera/PerspectiveCameraNode.swift
```diff
@@ -16,45 +16,78 @@ public class PerspectiveCameraNode : ObjectNode<PerspectiveCamera>
...
-        
+        // Position, scale and orientation carry the camera's own defaults and
+        // wording: an object's position default would put the camera on its
+        // subject, and its scale and orientation describe an object in a
+        // scene rather than the view of one.
+        let ports = super.registerPorts(context: context).filter { !["inputPosition", "inputScale", "inputOrientation"].contains($0.name) }
+
         return  [
-                    ("inputLookAt", ParameterPort(parameter:Float3Parameter("Look At", simd_float3(repeating:0), .inputfield, "Target position the camera points toward")) ),
+                    ("inputOrientation", ParameterPort(parameter:Float4Parameter("Orientation", defaultOrientation, .inputfield, "Which way the camera faces, as a quaternion (X, Y, Z, W). Identity looks along -Z")) ),
```
vade, line 57: Same concerns for Perspective as noted in Ortho.

vade (commented):
I'd like to understand the motivation of removing look at, which i found incredibly useful, and i find more helpful than orientation for a camera. 

If you feel strongly orientation is the way to go here - Is there a way we can replace that functionality, perhaps with a node that assigns a orientation for a position -> target vector (ray cast?). I found lookAt easy to use, feed in a target position that is animating (perhaps the animating position of a source of interest) and you can you can do whatever you want with camera position and and you always track. 

Now, theres a lot more work to do, and handling position and orientation together requires more of a dance, no? 

I also am not sure I understand the goal scale as a camera control. Help me understand the motivation there. My feeling is its not intuitive nor physical, but i might be misapprehending the goal. Its not quite a zoom as a i understand it (it isnt adjusting FOV).

tobyspark:
LookAt – one of LookAt and Orientation had to go, as one overwrote the other. We already have a better LookAt implementation in _Orientation Compose_ (has up), and the orientation input is definitive whereas the LookAt was not (had no up). The updated node description points at the compose node. See also: the commit comment.

Scale – You made a camera node with scale! As the PR states, the goal is to fix what we have, given #181. The danger here is we lose a set of clear wins that make the current camera system workable as this either turns into #181 mk.2, or eats into time we should be spending on e.g. serialisation.

Fabric/Nodes/Object/Camera/PerspectiveCameraNode.swift
tobyspark: see main conversation comment

Fabric/Nodes/Object/Camera/OrthographicCameraNode.swift
tobyspark: We don’t, but that’s a design discussion. See main conversation comment

Fabric/Nodes/Object/Camera/OrthographicCameraNode.swift
tobyspark: See main conversation comment.

Fabric/Graph/GraphRenderer.swift
tobyspark: No, but it could be clearer. The commit comment in 13b0270570dc8a6c7d3fb398ea87580a0f926914 explains it better. I’ll remove that.

vade:
Re Scale, ok, fair enough ha! I figured it was just due to how the node inheritance was set up, but Satin's camera has a scale so we might as well keep it for api continuity.

Just for the record - Orientation is not lookAt - lookAt gives you a specific point in world coordinates which is transformed based off of the objects world position and creates an orientation that centers the destination position on the ray. its a  ray cast from obj to obj. 

check out lookAt in Satin on Object

``` public func lookAt(target: simd_float3, up: simd_float3 = Satin.worldUpDirection, local: Bool = true) {
        if local {
            localMatrix = lookAtMatrix3f(position, target, up)
        } else {
            worldMatrix = lookAtMatrix3f(worldPosition, target, up)
        }
    }
```

Maybe we merge this, and add a look at node at some point, but not sure how we get those matrices. Maybe mesh / objects get output ports in the future? I figured we'd need something anyway for hit testing...

tobyspark:
Screenshot 2026-08-18 at 18 17 05

If that doesn’t functionally get us what you want, then yes there is a further discussion to be had. But one way or another what we had wasn’t tenable, so we need to lose one or the other. FWIW I had another read of #181 and I don’t think we’re far off landing somewhere. So def merge and pick that back up after release.

vade:
Well TIL I dont know what we've got :D

PR #327, opened 2026-08-20
Feature – Preview toggable between in-editor and separate window

tobyspark:
Adds an app-wide setting for the output to render in a window (as before) or in the editor itself, replacing the node canvas’s background gradient. Windowed output when full-screen loses the toolbar for fully full-fullscreen-ness. Default remains windowed output. Per-document was explored but proved cumbersome (see first commit).

Screenshot 2026-08-20 at 22 35 08


tobyspark:
...I think the default should be in-editor, but that’s something that shouldn't just be my call.

tobyspark:
Gah – Pinned Satin commit is now wrong. Fixing.

vade:
LGTM!

PR #335, opened 2026-09-08
Open BaseGeometryNode to external plugins

### LINKED ISSUE #334: Allow external plug-ins to subclass BaseGeometryNode
when creating `BodyMeshProviderNode` I noticed that codex had to subclass Node instead of `BaseGeometryNode`. 

Let me know if interested and I can provide a PR with the changes. 

Codex notes:

BaseGeometryNode is public, but not open, so a plug-in in another Swift module cannot subclass it.

As a result, BodyMeshProviderNode has to recreate standard geometry-node behavior, including:

- The Primitive input
- The Geometry output
- Conversion from the Primitive menu value to MTLPrimitiveType
- Applying primitive changes to the geometry
- Force-publishing a reused geometry instance
- Standard dirty-state publication behavior

The node must also keep this duplicated behavior aligned with future changes to BaseGeometryNode.

Proposed changes
- Make BaseGeometryNode an open class.
- Keep the existing open geometry property as the geometry supplied by subclasses.
- Add an open geometry-update hook that external plug-ins can implement.
- Have the base execute() method call that hook, apply Primitive changes, and publish the Geometry output when required.
- Expose primitive conversion publicly only if subclasses need to apply it themselves.

This would allow external plug-ins to implement only their specialized geometry-generation work while Fabric continues to own the standard geometry ports and publication lifecycle.

jvcleave:
Expose the geometry-node subclassing surface, add a throwing full-context update hook, and keep primitive handling and forced geometry publication in the base lifecycle. Cover the API through a client-style subclass test.

closes #334

### DIFF
--- a/Fabric/Nodes/Base/BaseGeometryNode.swift
+++ b/Fabric/Nodes/Base/BaseGeometryNode.swift
@@ -14,8 +14,8 @@ import Metal
 /// input, Geometry output, and publication lifecycle.
 ///
 /// External plug-ins can subclass this type, provide ``geometry``, and override
-/// ``updateGeometry(renderer:executionInfo:renderPassDescriptor:commandBuffer:)``
-/// to perform their specialized geometry work.
+/// ``execute(renderer:executionInfo:renderPassDescriptor:commandBuffer:)`` when
+/// they need specialized geometry work.
 open class BaseGeometryNode : Node
 {
     override open class var name:String { "Geometry" }
@@ -41,7 +41,7 @@ open class BaseGeometryNode : Node
         fatalError("Subclasses must override geometry")
     }
             
-    open func evaluate(geometry:Geometry, atTime:TimeInterval) -> Bool
+    public func evaluate(geometry:Geometry, atTime:TimeInterval) -> Bool
     {
         var shouldOutput = false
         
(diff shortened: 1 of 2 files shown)


Fabric/Nodes/Base/BaseGeometryNode.swift
```diff
@@ -55,9 +61,28 @@ public class BaseGeometryNode : Node
...
-    public override func execute(renderer:GraphRenderer, executionInfo:GraphExecutionInfo, renderPassDescriptor: MTLRenderPassDescriptor, commandBuffer: MTLCommandBuffer) throws
+    /// Updates the geometry for one execution pass and reports whether the
+    /// stable geometry instance needs to be published again.
+    ///
+    /// The default implementation preserves the existing Fabric geometry-node
+    /// behavior. Plug-ins can override this throwing hook when their update
+    /// requires the renderer, command buffer, or recoverable error reporting.
+    /// Call `super` so Primitive changes and standard dirty-state publication
+    /// remain active.
+    open func updateGeometry(renderer: GraphRenderer,
```
vade: Why do we need this if we have execute / evaluate?

vade (commented):
Just a question on why we need a new override point for the BaseGeoNode

Fabric/Nodes/Base/BaseGeometryNode.swift
jvcleave: I think it's because the existing evaluate(geometry:atTime:) method doesn't throw errors.

here is the codex take:

```
So the new override point is mainly a compatibility-friendly way to support throwing external geometry updates without rewriting all existing Fabric geometry nodes.
```

vade: execute throws now so if you want to do heavy lifting you can see if you need to evaluate using the evaluate method (or implement your own logic to see if inputs updated in such a way to recompute) and then test and use your own geo function inside of execute, which can throw. 

id rather not increase the surface area of APIs folks need to implement :)

vade: and honestly, evaluate is kind of a hack and likely shouldnt canonical API surface area :)

vade:
Thanks this looks good!

Issue #199, opened 2026-03-14
Workflow for determining Templated Sub Patching


vade:
During our first community meeting @tobyspark raised an interesting workflow idea which some other node based tools allow

User Story

* I want to make some Subpatch which is something like a template
* I want to be able to instantiate multiple copies of this template, and customize the input port values
* I want to be able to make changes to the template and have it propagate to the instances
* I want a mechanism to 'specialize' one specific instance, such that it can be changed without other template instances inheriting that change. For example, we customize logic in this one special case. 

Snippets vs Templates:

Snippets:

Snippet are 'by value'

* Snippets are useful graphs that are instantited into some other graph as is, without any live link
* Snippets can be altered ex post facto and do not change any prior snippets instantiated anywhere.
* Editing a snippet means opening a separate graph and making changes to it, and saving it.
* This only updates any NEW edits.

Templates:

Templates are 'by reference' 

* Templates are connected to instances 'somehow' 
* Changes in the template propagate to instances wherever they are instantiated (maybe across documents, but at a minimum within the same document / graph). 
* Templates can be disconnected from their instance (think 'deep copy') and create a new unique subgraph decoupled from their source Instance, and are efffectively now just a Subpatch. 

Change propagation 
* suppose i make a template, instantiate a few copies and connect published inputs and outputs
* i change the root template, and remove the published inputs and save the template.
* graph connections break as published ports no longer exist
* when i save the template, any instances of said template immediately detect the change through some 'mechanism' and update their nodes and published inputs and outputs. 

Questions:

* If I do not specialize a instance of a template / subgraph - is it editable?
* If its editable, what does that do? Specialize it and de-sync it from the template, or... ?
* When I edit the template, and im editing a separate graph / window, or am i somehow in a 'edit' of the subgraph inside of the main window. 
* How does undo behavior work?
* Can i have multiple documents with the same template loaded? If so, what happens when edits to the template happen?

Some ideas:

a) File backed Templates.

Toby pointed out this is cumbersome to 'ship' and handle file links, but lets discuss regardless. 

* Graphs are already supported to be nested and executed.
* Graphs are already supported to be saved on to disk.
* We could in theory make a file backed subpatch node which can 'watch' for changes and update its own subgraph accordinly

b) Use existing Subgraph Node  as entry point to templates

* Can we use the existing subgraph node, and have a special operation which implies that a subgraph is a live linked copy of another?
* User makes a subgraph that is standard today
[...]

PR #337, opened 2026-09-09
Feature – Node titles gain an icon

tobyspark:
Currently, Math Expression has an ad-hoc way of placing a warning icon into the node’s title when the expression fails to parse. This PR provides an API and the machinery to do this in a consistent way.

Screenshot 2026-09-09 at 18 21 42

This PR has the Math Expression node reworked to take advantage of this feature, as above.

This feature could be hooked up to display node execution failures etc. given the new throwable API that plugins brought. That is probably best for a separate PR.

The initiating work for this however lies in UI experimentation for #199

Screenshot 2026-09-09 at 10 49 45


vade:
so two minor concerns:

a) node is exposing icons - i think it should expose 'status' instead, and then the NodeModelView can map status to icons display. I know its nit picky and maybe a style thing, but i think that might allow for some more robustness down the line. Maybe node actually exposes an nillable error type? The gist of my nit is, the node is violating separation of responsibilities now - it should not be concerned with UI icons, but communicating what the icon represents. the view layer can then determine what icon to draw.

b) am i right in understanding that `.help(titleIcon?.tooltip ?? "")` installs an empty help value on every normal node? Can we make it such that all nodes dont get additional hover / help text unless present somehow? Not sure if this is a valid concern?

tobyspark:
A – At heart, you’re arguing for the icon to be a system-level feature, a limited set of meanings. As implemented, it’s open to a node (and so third parties beyond us) finding the provision of an icon useful, and doing something with it (that’s what was in my head when I made it so, largely because I knew with #199 a new use of it was coming). Consider those two quite different things, and for me at least the focus falls away from technical separation of responsibilities or suchlike, to be more about the app we want and community that can shape that. 

Now! I think the icon being a system-level feature is fine. Handling error and clone is sufficient for current purposes and visually clear. On not holding innovation back, Fabric is open-source, and if there are future ideas about what it could do, the system can be hacked, forked, whatever. Plus a truly open system would probably need to accommodate multiple icons, and then we’ve got Android.

The implementation is already part way there with the `tint` enum. I’ll rework along those lines. Plus sanity check what’s going on with empty `.help`.

vade:
Yea, I'd say if our statuses arent enough for a specific node author, we should update the statuses and thats a valid concern from an author, and we can update the engine to fix (and then be consistent for everyone else).

vade:
One other thought - we will soon have to content with

* Node's having a DynamicTimeBase UI - ie a way to choose which time base (ie provided or external)
* Async Execution - a node may free run some tasks like movie playback that can slightly drift (maybe ML detection) or Sync for rendering or a preview mode

Maybe not this PR, but it might be worth thinking about how we communicate those ideas in the UI? Im sure youve been percolating!

tobyspark:
A – Implemented. Note there’s precedence ordering, so errors, warnings and, in time, clone status can all be requested and shown in tooltip, but only the most important will be shown as icon.
B – Rather than further futzing with conditionals and SwiftUI identity, the tooltip is now on the icon rather than whole row, so it’s only present when there’s a status. Reduces the tooltip hit area, but the icon is where you’d hover if you want it.

Timebase – I think this isn’t a status, or a purely display icon, so probably should have nothing to do with this?
Async – I could imagine a status to show a long running task. Would need some heuristics about being informative but not too flashy, e.g. if tasks are short and/or regular.

vade:
Oh, yea i wasnt suggesting the time base and async be considered as icons, but i was sort of prompting your general UI/UX brain on the general concerns of having nodes having to attend to more signaling writ large (not in the scope of this pr) :)

vade:
My last piece of feedback here: We have FabricError which we should likely use / extend to cover any error cases not already covered (execution / syntax is there). 

My suspicion is we dont need a new type for Node Status?

tobyspark:
We do want a new type as the status should not be constrained to what FabricError should model. Concretely, near-future use like `cloned` / `linked` isn’t an error type.

vade:
Thats fair, i hadnt considered those having icons etc. Good call.

PR #338, opened 2026-09-10
Feature – NodeRegistry maintains a list of _Sub Graph_ type nodes

tobyspark:
...and with that, the “Embed Selection In...” context menu item also shows third-party nodes that are _Sub Graph_ subclasses. This list used to be hard-coded.

Screenshot 2026-09-10 at 08 09 42

### DIFF
--- a/Fabric/Nodes/NodeRegistry.swift
+++ b/Fabric/Nodes/NodeRegistry.swift
@@ -56,13 +56,39 @@ public class NodeRegistry
     }
 
     /// All UTTypes accepted by drop-target nodes, for use with drop destination handlers.
-    public lazy private(set) var allSupportedDropTypes: [UTType] = {
-        return self.nodeFileLoadingClasses.flatMap { $0.supportedContentTypes }
-    }()
+    public var allSupportedDropTypes: [UTType]
+    {
+        self.nodeFileLoadingClasses.flatMap { $0.supportedContentTypes }
+    }
 
-    public lazy private(set) var availableNodes: [NodeClassWrapper] = {
-        return PluginLoader.shared.pluginNodeWrappers
-    }()
+    /// Every registered node, read live from the plugin loader so a plugin
+    /// loaded after first access is included. PluginLoader.pluginsDidChange
+    /// says when to re-read.
+    public var availableNodes: [NodeClassWrapper]
+    {
+        PluginLoader.shared.pluginNodeWrappers
+    }
+
+    /// A registered node class that is a kind of subgraph, for choices such
+    /// as Embed Selection In.
+    public struct SubgraphNodeType: Identifiable
+    {
+        public let wrapper: NodeClassWrapper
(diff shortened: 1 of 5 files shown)


vade:
Im generally not a fan of this approach to the storing a list of available subgraphs. 

Why not just have a fixed list, calculated at loadPlugins, which is read only until plugins are loaded again (recomputed by NodeRegistry)

this avoids updates, provides a fixed array thats cached.

Frankly speaking every single part of this feels a completely over engineered and unnecessarily complex. Typical of LLM's gone wild :) 

Why not just:

```@MainActor
public final class NodeRegistry
{
    public private(set) var availableNodes: [NodeClassWrapper] = []
    public private(set) var subgraphNodeTypes: [SubgraphNodeType] = []
    public private(set) var supportedDropTypes: [UTType] = []

    public func loadPlugin(at url: URL) throws
    {
        let loadedNodes = try pluginLoader.loadPlugin(at: url)
        availableNodes.append(contentsOf: loadedNodes)
        rebuildDerivedLists()
    }

    private func rebuildDerivedLists()
    {
        subgraphNodeTypes = availableNodes.compactMap { wrapper in
            guard let nodeClass = wrapper.nodeClass as? SubgraphNode.Type else {
                return nil
            }

            return SubgraphNodeType(
                wrapper: wrapper,
                subgraphClass: nodeClass
            )
        }

        supportedDropTypes = availableNodes
            .compactMap { $0.nodeClass as? any NodeFileLoadingProtocol.Type }
            .flatMap(\.supportedContentTypes)
    }
}
```

and then key off of `subgraphNodeTypes` or some such? Am i missing something important about the list being dynamic?

tobyspark:
You are missing nothing. This is almost comic.

vade:
Hahah all good! I personally find I really have to fight the LLM to make the code not overly complex. And I really really really do not trust it to make good decisions unless - paranoia is its own reward!

tobyspark:
Haven’t reviewed via genie, but looks good to me and works in light testing.

vade:
Why do we need a new type `SubgraphNodeType` - could we not use a separate array of `NodeClassWrapper` objects and forgoe a new type?

tobyspark:
The reason was type narrowing, to not have to `as? SubgraphNode.Type` at all call sites. But there’s a middle-ground to be had that is consistent with `NodeClassWrapper` use. https://github.com/Fabric-Project/Fabric/pull/338/commits/7018721e97a852d1ce669b6b820cbea170f0df67

And if we’re simplifying, there’s no current need to dynamically load plugins subsequent to the initial load. e.g. The requirement for embedding software to provide its own nodes is handled with the main bundle lookup. We could extend that by adding a call to provide additional locations to search, but until we have anybody asking for it lets keep it simple. Empirically, _*spark stage_ is served as-is. https://github.com/Fabric-Project/Fabric/pull/338/commits/81e66b63d1b036f355902cfcd19b127fd312b7b5

PR #339, opened 2026-09-10
Feature – Sub Graphs can be cloned within documents

tobyspark:
An implementation for #199. The premise is that you can duplicate any subgraph type node as a clone, and then any edit within one of that _clone set_ changes them all. Key here is that this is _Fabric_ level functionality, not the introduction of a specialised node. For example, third-party [_SPK Stage Object_](https://tobyz.net/diary/2026/03/spark-stage-scenes/) nodes can be cloned.

This is the current new document spinning cube embedded in a _Sub Graph_ node. Now subgraph type nodes have _Duplicate as clone_ in their context menu –

Screenshot 2026-09-10 at 09 15 45

Now cloned, there are two spinning cubes. Note their execution is per-instance, and cloning operations maintain state on a running subgraph, so here we see the two cubes spinning out of sync with each other as the _Number Integrator_ has been running for a different amount of time in each. If this document were saved and opened afresh, those two integrator nodes would start integrating at the same time, and the cubes would be in sync. 

To control a cloned subgraph individually, publish ports. Here, we got the _Spin_ toggle and _Speed_ value from the new document graph for free, and e.g. toggling spin off on one will do what Fabric has taught you to expect: that node’s cube will stop spinning while the other node’s cube continues.

Screenshot 2026-09-10 at 09 16 52

This implementation likely can be improved but I think is the right direction. A simpler, more brutal approach of replacing each other members’ subgraph with the updated one fails to be simple in practice, requiring new subgraphs to be minted from the latest so that only shape not execution is cloned, and then any published ports to be remapped. Accepting some kind of change propagation machinery amongst subgraphs of any clone set, you arrive at either a peer-to-peer approach or a template-backed one. This implementation takes the template-backed approach, in practice serialising the desired subgraph shape to JSON. This has the advantage of being amenable to being backed by sidecar files or some kind of repository. This has the disadvantage documents saved with this will not degrade to uncloned duplicates if opened by earlier Fabric Editor, as the peer-to-peer model does. I don’t think that matters, and only having one copy of a cloned subgraph in a document is a filesize win. The clincher for me was it also allows this subgraph-scoped functionality to be contained to the Subgraph base type; the peer-to-peer implementation required clone tracking on the Node base type.

From the genie –

> A clone set is a group of Subgraph nodes kept identical in design while each executes on its own. Edit any member and the others follow. This is the template-backed implementation discussed on the issue: the canonical form of a set lives in the document as data, members carry a record relating them to it, and nothing clone-specific touches Node or Port.
> 
> UX
> 
> - Duplicate as Clone on any Subgraph node (Iterator, Deferred and Environment included) creates a sibling. The first clone starts a set named Set A, Set B, … Sets are renamed from the node menu; clearing the name restores a system one.
> - Members are peers. Enter any member and edit as normal: add, delete, wire, publish, rename, move, tweak a value, change a setting. Siblings update live, coalesced once edits pause for about a tenth of a second. Leaving a canvas and saving also sync.
> - Publishing is the specialisation boundary. Everything a node encodes replicates across the set, layout included. The one exception is the value of a published inlet, which the parent graph sets on each member through its proxy. To make a value differ per clone, publish it.
> - Runtime state is per member. Movie positions, feedback buffers, LFO phase and the like are never touched by a sync. A settings change that rebuilds a node's ports replaces that node in the siblings; everything else is updated in place.
> - Members are marked. A member wears the copies glyph at the trailing edge of its title, on the canvas and in the breadcrumb, shows the set name as its subtitle, and reads "Clone set: Set A, 3 members. Edits here reach every member." on hover.
> - Unlink from Clones turns a member back into an ordinary subgraph with its current contents. Sets nested inside it become sets of their own, so they detach from the former siblings but stay linked among themselves. Select Sibling Clones selects the set's other members on the current canvas.
> - Undo is per edit, on the member where it happened. A sync never appears as an undo step; undoing the edit syncs the siblings back.
> - A plain Duplicate of a member is an ordinary subgraph, not a clone. A set that would contain itself is refused.
> 
> Implementation
> 
> Model. CloneSet is an object on the document's root graph: id, name, the member node class, and the template as canonical JSON, the encoded form of a member's sub graph with every id rewritten to the template's own ids. It is saved with the document and dropped when no member refers to it. SubgraphNode carries cloneSetID and cloneRecord, a dictionary from template ids to the member's local ids for nodes, ports, wires and nested graphs. Node gains nothing.
> 
> Why the record is robust. Duplication rewrites every UUID in a node's JSON. The record's keys are template ids, which a rewrite never touches, and its values are local ids, which a rewrite maps along with the nodes. So Duplicate as Clone yields a correct record from the duplication itself, and a plain Duplicate strips the links. The record also means local ids stay unique per member, so the parent graph's wiring and any host lookup by port id are unaffected.
> 
> Sync. The edited member refreshes the template, then each sibling is reconciled from it through the two records: source local → template → sibling local. Surplus nodes go, missing ones are decoded from the source with ids mapped through the records, and matched nodes take layout, rename, published state and the values of inlets that are neither published nor wired. A node whose class or settings signature differs is replaced under the ids the sibling already recorded, so proxies keep their ids across the replacement. Wires are diffed by the sibling's own port ids; nested sub graphs reconcile first so the proxies the parent's wires land on exist. Nothing registers undo. The settings signature is the node's encoded form minus ideand clone links, with UUIDs normalised.
> 
> Template as source. One primitive makes a member from the template alone, in the set's member class, with every id fresh and recorded. It serves new members, recovery of a member with no record, which is rebuilt in place with the parent's wires re-bound to its published    ports by name, and a template updated from outside the a transient member. This is the path a sidecar or shared template would take.
> 
> Change detection. Graph gains a content revision, bumped by its own mutation API. While a node is a member it owns a CloneMemberObserverthat watches its node tree through the signals nodes, py publish: the offset and ports-changed subjects, thesubtitle subject read for the rename alone, observation tracking on a port's publish state, and each parameter's value publisher, counted only for an unwired, unpublished inlet. Every sink drops anything not on the main thread, since some subjects fire per frame from the render thread. A CloneSetCoordinator on the root graph debounces syncs and is main-thread state; calls from elsewhere are handed to the main actor.
> 
> Editor. Node menu items, a rename alert, breadcrumb entries showing a node's title icon, sync on leaving a canvas and a flush before    save. The glyph itself comes through Node.deriveTitleIcR.
> 
> One fix on the way. JSON parsed with JSONSerialization carries NSNumbers that the AnyCodable bridge re-encodes zeros as booleans, which would have corrupted saved templates. Clone paths parsead. A SuperShape parameter whose default sat below itsslider minimum, found by the fidelity test, is fixed as well.                                                                           
> Tests
> 
> 36 tests in CloneSetTests: records through duplication, save and unlink; every kind of edit reaching a sibling, including nested graphs and proxies; per-member published values; node replacemtion from the template, recovery and external templateapplication; live sync scheduling, debounce and undo; the badge and its view model mirror; and a fidelity pass that clones every core node type and expects a reconcile to change nothing.
> 
> Not in this PR
> 
> Named sets in the node library, per-node clone immunity, and sidecar or shared templates. The model is shaped for all three.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,13 +81,14 @@ For all development:
   - `class var name` — override: the stable name the type is registered and listed under, and the default instance title.
   - `func deriveTitle() -> String` — override only when an instance has a more specific authoritative title, such as a shader-backed `BaseImageNode`; defaults to `Self.name`.
   - `func deriveSubtitle() -> String?` — override where the node describes itself, nil otherwise: Math Expression's expression, StrategyNode's strategy.
+  - `func deriveTitleIcon() -> NodeTitleIcon?` — override where the node has something to flag at the trailing edge of its title row, with a tooltip: Math Expression's parse error. nil otherwise.
   - `var userName: String?` — final: the user's rename, serialized and public for rename-specific callers.
   - An empty name is no name: `userName` normalizes "" to nil at set and decode.
 - From those it composes, both final:
   - `var title` — the normalized result of `deriveTitle()`, falling back to `Self.name` when empty.
   - `var subtitle` — `userName ?? deriveSubtitle()`, normalized so empty values and values equal to `title` read as nil. General consumers read this rather than `userName`.
   - `var debugDescription` — the title plus resolved subtitle, e.g. `Math Expression (My Rename)`. `print(node)` and `"\(node)"` give it; diagnostic labels and traces use it when they need both values.
-- Fire `subtitleSubject.send()` whenever state feeding `deriveTitle()` or `deriveSubtitle()` changes; `NodeViewModel` mirrors the node's `title` / `subtitle` observably and keeps `userName` only for rename editing, clearing, and undo.
+- Fire `subtitleSubject.send()` whenever state feeding `deriveTitle()`, `deriveSubtitle()` or `deriveTitleIcon()` changes; `NodeViewModel` mirrors the node's `title` / `subtitle` / `titleIcon` observably and keeps `userName` only for rename editing, clearing, and undo.
 - Execution is **pull-based**; one execute per node per pass.
 - `GraphRenderer` (executor and scheduler) today does not use `nodeExecutionMode` or `nodeTimeMode` but will in the future.
 - **Iterator (QC-style)** remains the multi-evaluation macro; refinements allowed, paradigm fixed.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -171,6 +171,7 @@ Some nodes operate identically regardless of what data flows through them (e.g. 
 ### 3.4  Subgraph Behavior
 - Iterator applies per-iteration params before subgraph execute.
 - Render-to-Image-with-Depth sizes to inputs, attaches depth, outputs typed textures.
+- Clone sets (`CloneSet`, `Graph+CloneSet.swift`, `Graph+CloneReconcile.swift`): what a node encodes is its design and replicates across members; what it does not encode is runtime and stays per member. Published inlet values are the one encoded exception. Identity across members is the member's `cloneRecord` (template id → local id), never a field on Node; a settings change that rebuilds ports replaces the sibling's node under the ids it recorded, everything else reconciles in place. Edits reach siblings through `Graph.noteContentChanged()`: the Graph mutation API bumps it, and a member's `CloneMemberObserver` bumps it for the signals nodes, ports and parameters already publish. Nothing clone-specific belongs on Node or Port.
 
 ---
 
--- a/Fabric Editor/FabricDocument.swift
+++ b/Fabric Editor/FabricDocument.swift
@@ -336,7 +336,8 @@ class FabricDocument: FileDocument
     {
         let encoder = JSONEncoder()
         encoder.outputFormatting = [.prettyPrinted]
-        
+
+        self.editingContext.rootGraph.flushPendingCloneSync()
         let data = try encoder.encode(self.editingContext.rootGraph)
         
         return .init(regularFileWithContents: data)
--- a/Fabric Editor/Views/BreadcrumbEntry.swift
+++ b/Fabric Editor/Views/BreadcrumbEntry.swift
@@ -1,0 +1,39 @@
+//
+//  BreadcrumbEntry.swift
+//  Fabric Editor
+//
+//  Created by Claude on 9/9/26.
+//
+
+import SwiftUI
+import Fabric
+
+/// A breadcrumb entry: the subgraph node's title, preceded by its title icon
+/// where it has one (a clone set member's glyph), with the icon's tooltip on
+/// hover, the same way the node shows it on the canvas.
+struct BreadcrumbEntry: View
+{
+    let node: SubgraphNode
+    let nodeViewModel: NodeViewModel?
+    let action: () -> Void
+
+    var body: some View
+    {
+        let titleIcon = nodeViewModel?.titleIcon
+
+        Button(action: action)
+        {
+            HStack(spacing: 4)
+            {
+                if let titleIcon
+                {
+                    Image(systemName: titleIcon.systemName)
+                }
+                Text(node.title)
+            }
+        }
+        .font(.headline)
+        .buttonStyle(.plain)
+        .help(titleIcon?.tooltip ?? "")
+    }
+}
--- a/Fabric Editor/Views/ContentView.swift
+++ b/Fabric Editor/Views/ContentView.swift
@@ -76,9 +76,11 @@ struct ContentView: View {
                     ForEach(self.document.editingContext.entries) { node in
                         Text("›")
                             .font(.headline)
-                        Button(node.title) { self.document.editingContext.popTo(node) }
-                            .font(.headline)
-                            .buttonStyle(.plain)
+                        BreadcrumbEntry(node: node,
+                                        nodeViewModel: node.graph?.viewModelIfPresent(for: node))
+                        {
+                            self.document.editingContext.popTo(node)
+                        }
                     }
                 }
                 .padding(.horizontal)
--- a/Fabric/Graph/CloneSet.swift
+++ b/Fabric/Graph/CloneSet.swift
@@ -1,0 +1,77 @@
+//
+//  CloneSet.swift
+//  Fabric
+//
+//  Created by Claude on 9/10/26.
+//
+
+import Foundation
+internal import AnyCodable
+
+/// A clone set: the template every member is kept identical to, held as
+/// data, plus the set's name. Owned by the document's root graph and saved
+/// with it. Members refer to it by id and keep a record mapping the
+/// template's ids to their own; see SubgraphNode.cloneSetID.
+///
+/// The template is the encoded form of a member's sub graph with every id
+/// rewritten to the template's own ids. It is refreshed from whichever member
+/// was edited last, materialised when a member is created without a live
+/// source, and is the form that can be stored or shared outside the document.
+public final class CloneSet: Codable, Identifiable
+{
+    public let id: UUID
+    public internal(set) var name: String
+
+    /// Canonical JSON of the template graph, keys sorted, so two templates
+    /// with the same design compare equal byte for byte.
+    public internal(set) var templateJSON: Data
+
+    init(id: UUID = UUID(), name: String, templateJSON: Data)
+    {
+        self.id = id
+        self.name = name
+        self.templateJSON = templateJSON
+    }
+
+    /// The template as a JSON object, for rewriting ids.
+    var templateObject: [String: Any]
+    {
+        get
+        {
+            (try? JSONSerialization.jsonObject(with: templateJSON) as? [String: Any]) ?? [:]
+        }
+        set
+        {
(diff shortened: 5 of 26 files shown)


vade:
I did a bit of an automated review, and uncovered a bunch of cases this approach does not track or clone, and in principle cant do (re-entrancy, etc)

Apologies for the pasted LLM 'review' that below that enumerates issues caught. 

I did some musing about the problem with Codex, and it does seem that simply decoding instances of the same graph, conceptually fixes a bunch of the problems the cloning approach cant (or perhaps, currently cannot) - the only thing i need to vet is that strictly speaking subgraph a and b would have colloding UUIDs for subnodes. 

I dont think that is a problem. The goal of the UUIDs is not that for the entire document, every node needs a unique instance, but that every unique node has a unique id. The Clone having duplicated IDs is in fact a signal and semantically correct i would argue. Secondly, lookup and editing would do what you expect (each subgraph clone finds the same node, which is what it should do). And now since we dont have feedback cache, the only issue is if graph execution does anything dumb - but subgraphs are isolated from their parent graph as they are executed with their own graph renderer, so i think its ok. 

Secondly, decoding the same graph resolves the cloning problem as each decode instance is its own instance.

Overall this the cloning approach is super non trivial, has a lot of cases not accounted for, and potentially not possible to account for due to the flexibility of how some nodes can define their own internal state. 

My gut is syncronization is not the solution, but rather decoding each graph, and on edit, re-loading the graph and (but not the renderer or input times). This also resolves issues like out of sync during edit, in sync during fresh loading (ie the integrator nuance we spoke about)

In short, every part of my engineering brain says we should re-use the serialization / deserialization mechanism already available, that already works and not try to get fancy. 

The only issue is how we save these graphs, if we embed them in the root graph, or if we let them be sidecar files. 

I'll try to  propose an alternate design (not a PR, just a written proposal) later tonight. 

### LLM Review below:

```
[PR #339 — Cloned subgraphs](https://github.com/Fabric-Project/Fabric/pull/339)
Verdict: Do not merge in its current form.
This PR is much riskier than its tests suggest. It does not modify GraphRenderer directly, but it substantially modifies the central Graph model, mutation APIs, serialization, duplication, SubgraphNode, document saving, and automatically performs bulk graph reconciliation after edits.
Critical correctness findings
1. Multiple pending editors can silently lose the later edit.
CloneSetCoordinator keeps unique pending graphs in first-seen order. During flush(), it reconciles each graph sequentially.
If clone A is edited and then clone B is edited within the debounce window:
1. Pending order becomes [A, B].
2. A is applied to B, overwriting B’s newer change.
3. The already-overwritten B is applied back to A.
The implementation therefore produces “first editor wins,” despite the intended last-edited behavior. The coordinator must retain the latest source per clone set, with a revision/transaction ordering—not a list of graphs.
2. Many serialized node settings are never observed.
CloneMemberObserver only watches:
- node position,
- rename changes,
- portsChangedSubject,
- published-port state,
- backing parameter values.
That is not equivalent to “everything the node encodes.”
Concrete failures include:
- TimelineNode.duration and tracks, including keyframe edits.
- JavaScriptNode.scriptSource when the port interface does not change.
- JavaScriptNode.selectedExecutionMode and selectedTimeMode.
- Local LLM/VLM model selection, temperature, update interval, system prompt, chat mode, and context-token settings.
- Any plugin node with custom Codable settings that does not happen to emit one of the observer’s recognized signals.
These clones will silently diverge. Initial encode/decode fidelity tests cannot detect this; tests must mutate each category after cloning.
3. The SubgraphNode itself is excluded from the clone design.
The template serializes only member.subGraph, and the observer starts at member.subGraph.nodesRecursive(). The outer member node’s type-specific ports and settings are not part of synchronization.
Examples:
- Iterator iteration count.
- Environment texture/intensity.
- Deferred subgraph width/height.
- DeferredSubgraphNode.deferredMRTEnabled.
This is a fundamental mismatch between “cloned subgraph node” and what the template actually represents.
4. Invalid self-nested clone sets are allowed.
Reconciliation detects ancestor/descendant members and simply returns without synchronizing them. The test confirms synchronization is skipped, but that means the model has accepted an invalid clone set whose members cannot satisfy the clone invariant.
Creation, embedding, and movement APIs should reject the operation or explicitly unlink the affected member.
Significant risks
- Saving is not safely flushed. fileWrapper(configuration:) calls synchronous flushPendingCloneSync(), but an off-main flush() merely schedules a main-actor task and returns. Encoding can therefore start with stale state or concurrently with reconciliation.
- Reconciliation automatically replaces nodes, ports, and connections after a debounce, without an explicit renderer-safe mutation barrier. That creates execution/topology risk if rendering overlaps the bulk update.
- Synchronization is expensive and runs on the main thread: whole-graph JSON encoding, JSON parsing/remapping, repeated per-node settings encoding, linear port lookups, and teardown/recreation of all observer subscriptions for every member.
- duplicateAsClone is not transactional. Encoding failures can be silently skipped, and failure after attaching the original member can leave a one-member clone set behind.
- Many important failures become empty results or no-ops rather than typed errors, making the public API difficult to diagnose.
- Public APIs expose template JSON as opaque Data, leaking persistence machinery instead of expressing a clean domain operation.
- Every member still serializes its complete subGraph, while the root also serializes the canonical template and ID records. The implementation therefore adds storage rather than achieving true shared-template storage.
- The feature adds several strongly coupled models—CloneSet, coordinator, member observer, ID records, reconciliation context/report—and over 3,000 lines across 26 files. That is substantial model and subsystem expansion.
A cleaner direction
I would not try to repair this primarily by adding more observer subscriptions.
The cleaner architecture is an explicit shared subgraph definition:
- A clone set owns one canonical SubgraphDefinition.
- Members reference that definition.
- Per-instance runtime values/state are stored separately from design state.
- Editing enters an explicit definition-edit transaction.
- The renderer receives instance-safe materializations at a defined graph mutation boundary.
If that model is too large for the first version, a safer intermediate design would be:
- One general graph/node “design changed” transaction API used by every settings mutation.
- Monotonic revisions and one latest source per clone set.
- Main-actor isolation for all graph design mutations.
- Explicit renderer suspension/snapshot invalidation around reconciliation.
- Outer SubgraphNode settings included in the design.
- Typed, throwing duplication/reconciliation APIs.
- A narrowly supported initial node/settings set, rather than claiming universal Codable fidelity.
Testing: 87 clone-set, graph-execution, routing, and subgraph tests passed. The suite is substantial, but it tests initial cloning and direct reconciliation much better than real delayed edit observation, concurrent sibling edits, saving isolation, or renderer interaction.
```

tobyspark:
I hadn’t done an automated review, expecting that would be the first thing you did, and if with a different genie then even better. So, not surprised that a bunch of things came up, but a bit surprised on the take-down of the approach in general. 

I suspect there will still need to be a similar adversarial review on whatever you come up with, as e.g. "Per-instance runtime values/state are stored separately from design state” sounds like coming up with a whole new execution model. That sounds like a massive trade for what I read as issues that of the more dealbreaker-looking actually point more to shoring up Fabric’s data model than a fundamental wrong concept.

But mainly just happy we’re good with the UX and that’s locked!

vade:
Im not suggesting we do that, thats just one option

The thing I want to go with, which a separate convo seems to imply adversarially as much less work to implement, is simply encoding / decoding the graph and instantiating it when changes occur. The instance being edited wont need serialization loop - but other ones would. It also woldnt need to de-duplicate UUIDs as each graph is being exceuted by the sub graphs node unique graph renderer instance - and semantically they are clones, so duplicate IDs is semantically correct by my measure. 

So the issue then is simply storage - and timing the execution / serialization loop so things change in a predetermined way. 

This also scopes the logic to literally only be changes to Subgraph. Literally nothing else would need to change, if im pondering it correctly.

tobyspark:
> A simpler, more brutal approach of replacing each other members’ subgraph with the updated one fails to be simple in practice
– me

If there’s a way to make that not "fail to be simple" in practice, then that’s worth exploring. It does come with a real degradation to the UX as the state in all other clones will be lost. I’m not sure it's quantifiable how bad that is, will depend on what’s going on in the subgraph. A trivial example: you tweak something and all the movies except one restart.

Also I forgot to comment that the observation that the UUID being the same across each clone set sub graphs’s nodes could be a feature not a dealbreaker is a good one. That could also bring the peer-to-peer model back into play I think.

The other thing I forgot to call out was hell no on sidecar files being how this lands. I’m 100% that the core UX has to be seamless in-document. Import/export, a setting to back that in-memory clone graph with a sidecar file, any and all are quite conceivable but have to be additional in my view. (If we had bundle documents I might be persuaded)

tobyspark:
I’ve had another look, and there’ll be some commits coming. Note not as doubling-down, just the spirit of a better bake-off. On which note, I do think you should land on an implementation rather than just design doc. Maybe your genie is better than mine, but I find docs never quite survive contact with the reality.

> 3. The SubgraphNode itself is excluded from the clone design

Forgot to flag that this is deliberate not an omission. These ports are how you specialise each clone. To me the clearer model is that everything _inside_ the node is cloned, and the ports you see in the regular graph are just duplicated as expected and then behave normally (no magic).

PR #348, opened 2026-09-15
Feature – Tests for parameter port conformance

tobyspark:
Two tests for parameter port conformance
- Parameter Ports – Sweep every node for inlets that should be parameter-backed, and proxies that should match
- Sub Graph Proxies – Test that every proxy survives a save


tobyspark:
These tests currently fail. Fixes are trivial I think, but some sense checking first
- Should ports for transforms be parameter ports? We don’t have a UI for them.
- NumericVirtualPort stands out as not being a parameter port. Was that deliberate?
- The control `.Color` being backed by vec3 and vec4 is expressing a need we have, but seems to be a problem here. 

Here’s the genie’s analysis –

> 1. Twenty inlets that could be parameter-backed aren't
> 
> 17 are Transform. Seven declare it by hand (Decompose/Translate/Rotate/Scale/Transpose/Invert Transform, GeometryToTransformArray); ten are the noise nodes, whose "Coordinate Transform" comes from BaseImageNode.syncDynamicValuePortFromMaterialParameter, which goes out of its way to build a parameter-less value port and to delete any parameter-backed port it finds.
> 
> The library contradicts itself here, which is the thing to settle before any burn-down:
> 
> - simd_float4x4 conforms to DefaultParameterProviding; Node.parameterToPort maps float4x4 to a ParameterPort; NumericTypeAgnosticNode and PassThroughNode<simd_float4x4> ship transform parameter ports today.
> - PortType.portForType(from:) declines it — "we dont have a parameter UI for it (yet)" — and BaseImageNode enforces the same.
> - Both are right about the UI: ParameterGroupView.buildInputField has no Float4x4Parameter branch, so such a parameter renders as Text(param.label). The "Transform" patching node already shows that bare label today.
> 
> 3 are JavaScript's script-declared inlets (a, b, threshold) — not a declaration at all. JavaScriptNode builds ports from PortType.makeFreshPort, which never returns a ParameterPort whatever the type. Every node that builds inlets from a PortType at runtime is non-conforming by construction, and that factory is also the seam for making the whole invariant automatic.
> 
> 2. A saved Numeric Virtual proxy comes back bound to nothing
> 
> AnyPort.swift:52 — in the isProxyPort branch, .NumericVirtual is the one case that builds a plain NumericVirtualPort instead of a ProxyPort. rebuildProxyPorts then discards it (not ProxyPortProtocol) and makes a fresh proxy, which lands on the same id only because proxies adopt the inner port's UUID — but with the parent graph's published flag and publishedName reset. Publishing a numeric-virtual port one level further up does not survive a reload.
> 
> Verified by publishing and renaming a proxy, then round-tripping:
> 
> Color  (Image Mesh)                    published=true  publishedName=Renamed In Parent
> Float  (Directional Light)             published=true  publishedName=Renamed In Parent
> Image  (Spot Light)                    published=true  publishedName=Renamed In Parent
> Virtual (Dictionary Set Value For Key) published=true  publishedName=Renamed In Parent
> Numeric Virtual (Distance)             published=false publishedName=nil          ← lost
> 
> One line: make it ProxyPort<PortValue> like every other case — ProxyPort restores the declared type from the payload, so .NumericVirtual is preserved.
> 
> 3. A colorpicker on a non-float4 parameter makes the port lie about its type
> 
> ParameterPort.portType answers .Color for any .colorpicker/.colorpalette control regardless of the value type. Duo Tone's Dark Color and Light Color are Float3Parameters with a colorpicker, so they are ParameterPort<SIMD3<Float>> ports claiming the simd_float4 wire type. Two verified consequences:
> 
> - Proxy decode throws. A saved proxy of type .Color decodes as ProxyPort<simd_float4>, the inner port is a ParameterPort<simd_float3>, and ProxyPortBindingError.typeMismatch is raised. SubgraphNode catches it around the whole array, so all four of that node's proxies lose their parent-graph state, not just the two colours.
> - A colour wire into those inlets silently does nothing. canConnect returns true, the connection is made, and the value never arrives — the inlet keeps its previous value. Verified against hroughNode upstream: canConnect=true, connected=true,inlet SIMD3(0,1,1) before and after a red send.
> 
> Only 2 of the registry's 27 .Color ports are affected, both on Duo Tone. The fix I'd propose is at the cause — answer .Color only when the value type is simd_float4 — and the picker UI is unaffeView already has a Float3Parameter init. The cost isthat saved Duo Tone colour values stop restoring, as hydrate matches on portType.

vade:
Note there is a conceptual difference between parameter backed ports, parameter backed ports which expose UI, and standard ports.  A parameter backed port wires to a uniform buffer binding from satin and has combine / parameter group submission for rendering out of the box. Standard ports dont. Some parameters are mapped to UI.


> Should ports for transforms be parameter ports? We don’t have a UI for them.

No, i dont think so. IMO, the UI is hard to navigate or make intuitive for it no matter how you cut it IMO for matrix inputs. Theres at least a few variants and none of them cover all cases or express all values nicely other than 16 numerical inputs which i think is gross and not friendly at all. I do not think transforms should be decomposed to a matrix of inputs, or a special UI in my mind. At least im not seeing a good solution at first and second glance.

> NumericVirtualPort stands out as not being a parameter port. Was that deliberate?

Yes, a Numeric means it can have math done on it, not that its a  singular number, ie Quaternion, Vector, Transform, its virtual but implies can be operated on like math, as opposed to a String or Reference type like Geo or Mat. See `NumericValueOperations` for example.  Its a virtual node that allows numeric type casting to happen independent of other type casting operations. 


> The control .Color being backed by vec3 and vec4 is expressing a need we have, but seems to be a problem here.

I dont see the problem in principle, only that we need to make typecast happen which we havent quite gotten to? 


Genie feedback:

>The library contradicts itself here, which is the thing to settle before any burn-down:
 
 
I think if we differentiate parameter not requiring UI, its not an issue? (see my first note).

vade:
As for the proxy port fix for numeric, that seems like a good catch and a good easy fix !

vade:
And note, while we could make a Transform expose 3 vectors for Rotate Transform Scale, as an input port it may have a shear matrix which cant be expressed or edited with those controls, so its not quite so clean (I decided to revisit my assumption about the UI and i still think no UI for Transform?)

tobyspark:
Javascript node and NumericVirtual proxy fixes pushed.

On colour, I can’t see a way around either splitting the `.Color` controlType into RGB and RGBA variants, or adding a `color` value type to Fabric. There’s some voodoo that could be done to introspect, but nothing gets around 

```
        case .Color:
            return simd.simd_float4.self // could this backire?
```

On transforms, I’m confused. I agree there’s no sane UI to present, so it’s not user editable. That rationale of parameter port is out. The other rationale, of always being backed by a value, confuses me. To my understanding, the rationale that has us here is we’ve imported a concept from Satin that provides a default value that is required for the object to execute. Expressed like that, it feels like an internal implementation detail of the node, and shouldn’t leak into the UX. But I might be wrong, and given how closely bound Fabric and Satin are, there are likely other costs to pay if it weren’t that way.

PR #350, opened 2026-09-16
Feature – Syphon client node robustness

tobyspark:
The Syphon client would connect in interactive use if the Syphon source was already there, but other sequencing would fail, e.g. Syphon source start after node creation, or Syphon source restart. There also was no way to tell the node to disconnect (short of asking it to connect to an unlikely name). This fixes that. Prompted by use in *spark stage.


tobyspark:
Aaaaaand a new commit, because image flipped-ness wasn’t correct. Happily for Fabric before, the incorrect behaviour cancelled out, so it took a third-party server/client to see.

vade:
Did an auto LLM review - i need to validate some of this, but heres some feedback in the interim:

Verdict: request changes. The reconnect idea is sound, but the orientation commit is backwards and currently masks itself in Fabric-to-Fabric testing.

Findings

1. [P1] The Syphon flip contract is reversed.
   Fabric’s textureTransform maps canonical presentation UVs to stored texture UVs:
   - identity: storage is already canonical
   - .textureVerticalFlip: storage must be flipped when sampled
   Syphon’s flipped: parameter describes the input texture: YES means the supplied texture is vertically flipped. Its Metal implementation directly blits when false and redraws with a flip when true. Therefore the correct predicate is the existing one:
   inputImage.textureTransform == .textureVerticalFlip
   The PR changes this to !=, meaning canonical identity images are flipped unnecessarily, while actually flipped images are not corrected. It also incorrectly marks every received Syphon texture as vertically flipped. Fabric client → Fabric server appears correct because the two mistakes cancel; interoperability with a conforming third-party endpoint will be inverted.
   This is confirmed by:
   - PR server implementation
   - PR client transform
   - Syphon’s flipped: API contract
   - Syphon’s false/direct-blit implementation
   - Old QC server forwarding textureFlipped
   - Old QC client publishing its copied output as flipped:NO
   Arbitrary transforms are also mishandled: a rotated or cropped movie transform is treated as “flip required.” The server must either normalize such images first or explicitly support only identity and vertical flip.

2. [P2] A crash/relaunch can remain bound to the stale server.
   The PR states that a dead client can continue reporting isValid, but its recovery logic only blacklists the bound identity after isValid becomes false. When a relaunched server is appended alongside the stale entry, matches.first selects the stale entry again; because its identity equals boundServerIdentity, no new client is created.
   This persists until Syphon’s directory eventually expires the stale entry. The old name-bound implementation deliberately selects lastObject, i.e. the newest matching announcement.
   See the PR’s first-match selection. Recovery should use notification identity/state explicitly or otherwise prefer the newly announced matching UUID without disrupting intentionally duplicated servers.
3. [P2] The new observers are not tied to Fabric’s execution lifecycle.
   Directory observation starts in both initializers and ends only in deinit. Stopping rendering leaves the observers active and may leave the Syphon client connected until the graph releases the node. The QC plug-in creates these resources in startExecution and releases them in stopExecution.
   Register observers and force an initial lookup in startExecution; remove observers, stop/nil the client, clear the output, and reset binding state in stopExecution.
4. [P2] The orientation tests assert the implementation’s assumption, not the protocol.
   Both tests call the new helper and encode the same reversed premise, so they pass while interoperability is broken. Use an asymmetric pixel fixture and verify actual stored/output rows—or at minimum assert Syphon’s documented input contract:
   - identity → flipped: false
   - .textureVerticalFlip → flipped: true
  
Performance parity
The old QC client only produces a new output when hasNewFrame is true. Fabric currently creates a new FabricImage UUID and sends it every execution even when Syphon returns the same underlying texture. That dirties downstream nodes and allocates continuously. Cache the current FabricImage, update/send it only for a new frame, and preserve the last image until the client disconnects.

The exact PR builds, and its two orientation tests pass, but they validate the incorrect model. GitHub reports no CI checks for the branch. I made no source changes to the working tree.

vade:
Quick human assessment - LLMs cant reason about flippedness lol.

tobyspark:
If you pointed your genie at the source of the QC plugin, that was a good call.

> [P1] The Syphon flip contract is reversed.

Questionable reasoning aside be good to have you check this. Empirically *spark stage's non-Fabric Syphon output has the correct output paired with Syphon recorder but not with Fabric as-was.

> [P2] A crash/relaunch can remain bound to the stale server.

I think the reasoning and conclusion is wrong. My understanding is Syphon will mint a new UUID for each connection. More importantly, if I force-quit *spark stage and relaunch, Fabric picks up the new connection. 

> [P2] The new observers are not tied to Fabric’s execution lifecycle.

That’s an obvious miss. I’ll make it so.

> [P2] The orientation tests assert the implementation’s assumption, not the protocol.

If that’s suggesting making a Syphon server and client to test against... maybe? Aside from the overhead, isn’t that going to be just as circular logic?

tobyspark:
Fixes pushed. Includes a wholly different approach to Syphon Server observing: one process-wide directory. The client node has also gained a settings view where you can set the input from a list of current servers.

Screenshot 2026-09-21 at 22 56 33

vade:
May i suggest following the pattern of the Syphon QC plugin?  A Syphon Server Directory Node which pushes outputs when server availability changes, allowing use of other patches like sample and hold or what not to re-assign / select servers to pass to the Client? Its far more composable.

tobyspark:
We could certainly add that node. I think it’s a separate use-case to the settings view here though. If all you want to do is set the name input to the correct string, anything other than that settings view is a faff. Or actively not what you want, if it’s then dynamic.

PR #352, opened 2026-09-17
Fix – Camera selection when Sub Graph present

tobyspark:
A camera in a graph was silently ignored if the graph had a Sub Graph node. Fixed. 

Follow-up fixes: the graph is now correctly searched for camera nodes, and related dead code clean up.

Note the decision of #181 to have a single camera will likely obsolete the approach fixed here.


vade:
Hey, so this approach has some performance implications due to the recursive rescan:

The re-entrant GraphRenderer.execute diagnosis is correct, but I’m concerned that this solution moves recursive camera discovery onto the per-frame render path. Walking the graph and nested subgraphs—and creating intermediate arrays with filter, compactMap, and map—on every outer execution adds work proportional to graph size even though camera topology rarely changes.

Camera discovery and re-entrant execution should be treated as separate problems: cache the selected camera when graph topology changes, propagate invalidation when a nested graph’s camera changes, and allow only the outermost execute call to assign currentCamera. That preserves correctness for camera insertion/removal and nested graphs while keeping steady-state rendering O(1) for camera selection.

Ive got an alternative fix/ proposal coming shortly.

tobyspark:
FWIW, my aim was correctness knowing #181 would make all this redundant. And typical cases where the camera is in the root graph wouldn’t recurse.

Superceded by #352

Issue #349, opened 2026-09-16
UI Nit - Math Expression Code Box isnt styled like others.


vade:
We should ensure it inherits light / dark based on system setting.

tobyspark:
FYI I’ve got a whole bunch of live code node improvements coming. Prompted by hitting a few bugs in the Javascript node, expanded to the review I always intended to do, and now including this nit and general aligning the three live code nodes.

tobyspark:
Fixed in https://github.com/Fabric-Project/Fabric/pull/354

PR #354, opened 2026-09-19
Feature – Live code node improvements

tobyspark:
A bunch of fixes, alignments between, and improvements of, the live code nodes _Javascript_, _Math Expression_ and _Live Image_.

The history is I hit a bug or two using the Javascript node for the first time on an actual task. Then #349 got me looking across the live code nodes. And I always meant to do a proper review on the Javascript node, given the emphasis I put on Fabric having something like it early on.

> # Live code node improvements
> 
> Fabric has three live-coded nodes: **JavaScript**, **Math Expression** and **Live Image**. They had grown up separately, and it showed — three editors that looked different, two of them ignoring the system appearance, one language with no syntax colouring at all, an anachronism of a signature syntax, and errors that only existed inside a settings panel. This branch brings them to a common standard, and fixes what a review of it then turned up.
> 
> ## A shared editor
> 
> - **One appearance, applied once.** Theme, font and layout had been settled three times and settled differently — two editors asked for a fixed dark theme, one asked for nothing and got the package's light default; layout was an initializer argument in two and an environment value in the third. A single `codeEditorChrome()` modifier now carries all of it. Size deliberately stays with each editor, since only their placement knows how much room they get. (`270f82fd`)
> - **The system's appearance, not a hardcoded one.** `Theme.v(for:)` resolves against a colour scheme and each editor reads its own — `CodeEditor` takes a theme rather than reading the environment, so resolution happens at each of the three rather than once around them. Math Expression picks up the shared editor font on the way, which it previously had no way to have: the font travels with the theme, and it had no theme. (`9bb86820`)
> - **Each editor says what to write in it.** Math Expression already told you what an expression was for; the other two told you nothing, which is least affordable where the editor opens on a signature whose annotations name types you cannot see. One shared view now carries a summary and a guide link for all three. (`270f82fd`)
> 
> ## Math Expression gains a language
> 
> The editor was plain text — a keyword, a function, a number and a name all looked alike, in the one live-coded node whose editor could not tell you anything about what you had typed. It now has both halves, as JavaScript does: a `LanguageConfiguration` for what the lexer sees, and a completion service for what only the expression knows.
> 
> Neither is written in Fabric. The engine publishes the words it answers to, derived from the tables and types it works from rather than a list kept beside them, and `MathExpressionEditorSupport` turns those into a configuration and a completion set. A function, type, keyword or operator added to the language colours and completes here with nothing to change in Fabric. Written in Fabric, the configuration had *already* drifted — an operator regex that no longer matched the operator list beside it — and nothing could have caught it, because the two halves lived in different repositories. (`2eb8cb65`)
> 
> ## The JavaScript signature is TypeScript
> 
> The signature was a shape of its own — `function (__type name) main(__type name)`, outputs before the function's own name, types spelled as nothing else spells them. It is TypeScript now: parameters are the inputs, the return type is the outputs, types are named for what they are. `FabricTransform[]` and `Record<string, FabricTransform>` compose rather than enumerate, so the thirty-odd names the old form needed become thirteen plus two rules. Type names are offered as completions — most of what the annotation prefix was doing.
> 
> The old form still parses: a saved script belongs to a document, and a change of syntax cannot be what costs it its ports. What is read in the old form is written back in the new one, touching the signature alone and leaving the body as its author left it. (`1a34387b`)
> 
> **A guide, not just a signature.** A settings line cannot carry what a `FabricTransform` arrives as, that a quaternion's `w` comes last, that geometry and image are handles a script can pass on but not make, or that a `context` object is in scope at all. `JavaScriptNodeGuidance.markdown` carries it and the settings view links to it. (`d612c0a0`)
> 
> **The Ports box is gone.** It listed what the signature above it had just declared, in a second notation, and no other live-coded node has one. The signature is the statement and the node is the result; a third telling in between is a place for them to disagree. (`846666ff`)
> 
> ## Nodes report their own errors
> 
> - **An exception reaches the author.** A script that threw reported "must return an object containing the declared output values" at L1:C1, whatever had gone wrong and wherever — three separate faults between the exception and the panel, including a handler writing to a local that was never read again. The node now reports JavaScriptCore's own diagnostic, at the line and column it gave. (`8c301347`)
> - **Errors show on the canvas.** A script that will not compile or threw last frame, and a shader whose pipeline will not build, were known only inside a settings panel. Both nodes now derive the status the title row draws. (`d239fd09`)
> - **The reported line is the right line.** The signature pattern took the whitespace before the body brace into the match, so transpiling moved the body up a line — and multi-line parameter lists by more. That hit *every* migrated script, since the rewrite from the annotated form produces exactly that shape. The whitespace is now outside the match and the replacement is padded to the same line span, for the canonical form as well as the transpiled one. (`58f313bc`)
> - **An output that will not box says so.** Boxing failed for two different reasons and the node treated them alike: it sent nil and said nothing. Declining to set an output is still silent; a returned value that disagrees with its declared type now records a warning naming the output, its type, and what came back. (`6826f6ae`)
> - **A recovered node stops showing the error.** Live Image recorded a workspace error nothing cleared outside `init`, so one transient write failure parked the node for the life of the document. (`d05ae377`)
> 
> ## An edit asks to run again
> 
> `GraphRenderer` skips a `Processor` whose `isDirty` is false, and an edit is the one change a node makes to itself — nothing upstream moves, so nothing else marks it. Both nodes went on producing the *previous* program's output until something unrelated in the graph happened to run them; pasting a corrected script over a wrong one looked like the paste had not taken. `markDirty()` now happens on recompile, whichever path built the new runtime. (`93557ec4`, `dbb02bd6`)
> 
> Math Expression had already named this gap: `needsEvaluation` exists to "force one evaluation after a (re)compile even if no input changed" — the right intent applied one step too late, since it is read *inside* `execute`.
> 
> ## Also from review
> 
> - **`FabricValue` cannot be an output.** It earns its place on the way in — dictionary connection compatibility is exact type equality, and both the JSON parser and the dictionary nodes' default emit a dictionary of anything, so `Record<string, FabricValue>` is the only spelling that receives them. On the way out there is nothing to box into, and a script declaring one compiled, got a port, and emitted nothing from it every frame. Now refused where the signature is read, at any depth, in both syntaxes. (`7d5d1f85`)
> - **Diagnostics cross threads.** The graph executes on its own queue; deriving the node's status made the main thread read the diagnostics array every frame for every JavaScript node on the canvas. The array now lives in a `Mutex`, and `JavaScriptNodeDiagnostic` — public, so it got no implicit conformance — declares `Sendable`. (`d51f1fe8`)
> - **The editor keeps the rewritten script.** Canonicalisation happened on the node and the settings editor never heard, so the next keystroke wrote the annotated form back over it. (`98705576`)
> - **The theme is built once.** `Theme` re-stamps its `id` on any property write and `CodeEditor` restyles the whole text storage when that `id` changes — so a theme built inside a view modifier's body was restyling the editor on every keystroke. (`e57ec07a`)
> - **No unhandled-file warning.** The guidance markdown is named in `exclude` rather than moved, since the settings view links to it at its path on GitHub. (`7976cdba`)
> 
>


Fabric/Nodes/Image/LiveCode/LiveImageNode.swift
```diff
@@ -119,13 +122,32 @@ public class LiveImageNode: BaseImageNode
         do {
             // Use non-atomic writes to preserve inode/watcher continuity for live recompiles.
             try source.write(to: shaderFileURL, atomically: false, encoding: .utf8)
+            // A write that succeeds retires whatever the last one failed with.
+            // Nothing else clears this outside init, so without it one transient
+            // failure leaves the node refusing to run for the document's life.
+            self.workspaceError = nil
```
vade, line 128: clean. like it.

Fabric/Nodes/Parameters/Number/MathExpressionNode.swift
```diff
@@ -42,13 +44,9 @@ struct MathExpressionView: View
...
-            Text("Write an expression — free names become input ports, results become output ports. Beyond numbers, values can be vectors, transforms or arrays, and a single expression can drive several named inputs and outputs at once.")
-                .font(.caption)
-                .foregroundStyle(.secondary)
-                .fixedSize(horizontal: false, vertical: true)
-
-            Text("[Language guide ↗](https://github.com/tobyspark/MathExpressionEngine/blob/main/GUIDE.md)")
-                .font(.caption)
+            CodeEditorGuidance(
+                "Write an expression — free names become input ports, results become output ports. Beyond numbers, values can be vectors, transforms or arrays, and a single expression can drive several named inputs and outputs at once.",
+                guide: "[Language guide ↗](https://github.com/tobyspark/MathExpressionEngine/blob/main/GUIDE.md)")
```
vade, line 49: mind linking to Fabric as this is a canonical shipped node?

Fabric/Nodes/Utility/JavaScriptNode.swift
```diff
@@ -397,15 +413,56 @@ private final class JavaScriptNodeRuntime
...
 
+    /// What came back, from JavaScript's side of the bridge. An array's count is
+    /// worth carrying: a `FabricVector3` that arrives two long is the common
+    /// version of this mistake, and the count is the whole of the explanation.
+    private static func describe(_ value: JSValue) -> String
+    {
+        if value.isBoolean { return "a boolean" }
+        if value.isNumber { return "a number" }
+        if value.isString { return "a string" }
+        if value.isArray
```
vade, line 457: naive JS question, is dictionary handled by object, or some other type?

Fabric/Nodes/Utility/JavaScriptNode.swift
```diff
@@ -628,7 +731,7 @@ public final class JavaScriptNode: Node
     private static func defaultScriptSource() -> String
     {
         """
-        function (__number sum, __bool thresholdPassed) main(__number a, __number b, __number threshold) {
+        function main(a: FabricNumber, b: FabricNumber, threshold: FabricNumber): { sum: FabricNumber, thresholdPassed: FabricBool } {
```
vade, line 734: Im playing catch up to these are new types that a boxed and mirrored?

Im not a huge fan of the Fabric prefix? Can we make this feel more native and elegant if we are working on it?

vade (commented):
I havent run it yet, overall just have minor nits, largest concern is the naming convention of the bridged types. 

Suggest we drop the Fabric suffix, and just do Number, Bool, Int, etc? 

I saw the huge diff on the parsing, and im just simply not equipped to handle it. I'll do a robo-review of it over the weekend when i have more bandwidth and have the context of the changes myself now. 

Thanks!

Fabric/Nodes/Utility/JavaScriptNode.swift
tobyspark: We can lose the prefix. I hadn’t thought that could be a way of signalling opaque variants. I just thought it might be good to signal that there’s a type mapping going on. I’ll update.

Fabric/Nodes/Utility/JavaScriptNode.swift
tobyspark: Object. And, fun fact, an Array is an Object too!

Fabric/Nodes/Parameters/Number/MathExpressionNode.swift
tobyspark: Whoops that’s a regression. Will update.

vade:
Human / LLM Review:

Request changes. PR #354 has good intentions and thoughtful tests, but it is not mergeable yet.
Blocking findings

1. [P1] A newly created JavaScript node is broken by default.
   The parser now accepts Number and Bool, but defaultScriptSource() still emits FabricNumber and FabricBool.
   Result: every new node reports Unsupported Fabric type 'FabricNumber' and has zero input/output ports.
   Fix the template and add a test asserting a default node has no diagnostics and exposes its expected ports.
   
2. [P1] An invalid edit immediately runs the previous script and erases the compile error.
   compileAndSynchronizePorts() marks the node dirty, but its failure path leaves compiledSignature and runtime pointing at the last valid script. The next render executes that stale runtime, then replaces the compile diagnostic with its empty runtime diagnostics.
   The editor can therefore contain invalid code while the node looks healthy and continues emitting results from code the user no longer sees.
   Preserve the last-good ports if desired, but invalidate execution state on compilation failure or keep compilation diagnostics separate and refuse to execute stale code.

3. [P1] The checked-in dependency resolution contradicts Package.swift.
   Package.swift now requires MathExpressionEngine 1.2.0 and imports MathExpressionEditorSupport, but the checked-in Xcode Package.resolved still pins 1.1.1. That version does not publish MathExpressionEditorSupport.
   Automatic resolution may silently repair a developer’s checkout, but deterministic Xcode/CI builds with automatic resolution disabled will fail. Commit the 1.2.0 resolution.

4. [P2] “No output” scripts parse successfully but cannot execute.
   The new parser deliberately accepts a signature without a return type and includes a parser test for it. However, the runtime still unconditionally requires main to return an object.
   Consequently, this advertised consumer-style script always throws:
   function main(value: Number) {
     console.log(value)
   }
   When signature.outputs.isEmpty, an undefined return should be valid.
   
   we also likely 

5. [P2] The “top-level” signature regex parses signatures inside comments.
   firstMatch is applied directly to the entire source, without excluding comments or strings.
   A normal editing pattern such as this breaks compilation:
   // function main(oldValue: Number): { oldResult: Number } {
   function main(value: Number): { result: Number } {
     return { result: value }
   }
   It transpiles the commented signature and leaves the real TypeScript signature untouched. The parser needs lexical awareness of comments/strings rather than a raw first regex match.

Verification

The PR’s targeted 38 tests pass, but four isolated review probes covering the paths above all failed:
- Default node validity
- No-output execution
- Invalid-edit diagnostic persistence
- Ignoring commented signatures


additionally we might be able to make producer, consumer, processor switch auto

The JavaScript node does not infer its execution role from its ports.
Its role is manually selected in the settings panel:
- Defaults to .Processor
- Serialized as selectedExecutionMode
- User can choose Provider, Processor, or Consumer
- Rebuilding dynamic ports does not update the role
So a node with only outputs can remain a Processor, and one with only inputs can also remain a Processor unless manually changed.
Conceptually, automatic inference could map:
- Outputs only → Provider
- Inputs and outputs → Processor
- Inputs only → Consumer

But that logic does not exist.

tobyspark:
These commits address those issues. 

I like automatically determining the execution mode from the function signature. Note the preceeding-this-work manual version was partly broken – changing the selection did not notify the graph of the change, which this now does.

The parser was getting a bit ridiculous for something that more or less has to just skip comment lines and then extract a single line starting with `function`, so I’ve pushed back there for simplicity over catching some edge cases. 

Note from Live Image node out, we’ve established the idiom that these nodes throw on compile error. The alternative is to keep the node processing with the last valid script. I think throwing is the cleaner for now, but we might want to revisit down the line.

vade:
re-review passes, with the only nit being that the JS comment lexing is not perfect, but thats a nice to have and not a blocker. Merging!

PR #355, opened 2026-09-19
Fix – A string parameter is edited against all of itself, not the first hundred characters

tobyspark:
Try touching a bunch of JSON in a String value node, and you’ll hit this –

> The string input field read its parameter as prefix(100). That read is not only what the field shows: ParameterObservableModel compares it against the edited value to decide whether to write back, so it is the baseline every edit is measured from. Opening a node's settings a second time built a second field, seeded it with the first hundred characters of a value that was longer, and the next keystroke found it different from the same hundred characters and wrote the pair back — replacing the parameter with what had been shown. A pasted-in value long enough to need the truncation was exactly the value it destroyed.
> 
> The field reads the whole value. A single-line field still shows what fits and scrolls the rest, which is what it did with the first hundred characters anyway; what changes is that the rest is still there afterwards.


vade:
So this was there for a reason - huge UI snags when editing very large JSON or long multi line text files poduced huge UI lags directly related to the string field. 

I dont have a great suggestion TBH - as input fields should in theory show the full value (i'll grant you that) and multi line text editor inside of the node ui doesnt feel great to me, never the less, large strings should be handled gracefully somehow, i dont have good solutions and having a special 'large text editor' seems weird?  

Did you test this for very large inputs and note any lag? It was a huge issue in earlier testing. 

I'll try to give it run through over the weekend if i dont hear anything.  Try a very large text file -> string output -> string input somewhere and select the node, that KILLED things for me :(

tobyspark:
I figured it must be like that for a reason. Equally, outright data loss from under the fingers of the user is unacceptable, right?

Trying to pare apart the issue, there is the UI for handling large strings, and there is performance. 

On the UI, editing a stupid-long string in a text field sized to match number fields etc. is clearly not fit for purpose. What to do about that is less obvious. Perhaps at some size it just becomes read-only and you e.g. copy and paste in and out. Beyond that I _am_ open to there being different UI for different kinds/sizes of strings. To take my use that got us here a lot further: what if it knew it was JSON, and then pretty printed it with all values directly editable using appropriate widgets? [^1] I think that could actually be really neat, could work as a kind of settings node for your whole graph. So I suggest we should be open to the design space here, and not reject outright.

On performance, I don’t know enough about SwiftUI to have a view on it here beyond talking to a genie, and thinking fuck sake this was a solved problem decades ago. I didn’t notice anything last night, but had a lot of things in flight.

[^1]: Without wishing to bait anyone, in my mind one thing that TD got right is the idiom of dumping data into a node, to have it visibly on display and editable in the context of the machinery. I can’t remember any specifics well enough, beyond there being some kind of tabular data node, and the data-in and visualisation of that data being the same thing, or if not the same thing then having some coherence between them.

Issue #134, opened 2025-12-03
Relative vs Absolute file paths for Fabric files / Package files prep


vade:
given that we want to enable package files, a pre-requisite would be to enable relative file paths.

this opens some questions

1) Do we always save relative file paths?
2) If yes, side effects?, if not, how does a user intercede to choose?
3) If a relative path is opened on a system which has a pref for absolute, do we convert on save? Side effects?

Hrm.

vade:
Prep for #27

jvcleave:
easiest pattern i use is to create a folder that can be zipped up

<img width="237" height="115" alt="Image" src="https://github.com/user-attachments/assets/5bb58119-7277-4188-bb68-43d55099fe59" />

JustinMeans:
> easiest pattern i use is to create a folder that can be zipped up
> 
[...]

PR #365, opened 2026-09-21
Feature/document bundle

### LINKED ISSUE #27: Document Bundles for distributing assets as well as fabric graphs
Today a Fabric document consists just of the graph as a json representation.

Thats fine, but we might want to think about packaging assets within a Document Package (similar to work i did on Hideout) allowing for documents to be shared, have help or single download tutorial files as zips, etc.

Workflow

1.  Make a document in Fabric
2. Export document to a Fabric Bundle or some name
3. Exported file 'collects' and embeds all required files / assets

This would require some way of knowing what an asset is. 

We would need to iterate over every node, every param, and find a non nil / empty `StringParameter` with a `.filepicker` type, and copy the files into the destination bundle.

vade:
This PR adds

* Relative file support to NodeFileLoadingProtocol - and node implementations. On a saved Fabric doc, you can edit a file path to make it relative, and it will still work.

* Document Bundle Support - saving a graph to a Fabric Bundle allows us to identify all file references, gather them (copy) into the bundle, taking care to not copy duplicates (so they share references) and allowing a fabric document and its assets to move as a single file as seen by the system. 

this resolves #27 and #134


tobyspark:
- new document, drag in image, hook up to image mesh. see it, great.
- save. aha! now choice of format in dialog. select bundle.
- finder. show package contents. voilà, an assets folder containing that image.
- close document. open document. image node still has original, absolute path.

So what is the idea here? I expected that path to have changed to something like `<document assets token>/name.ext`.

Also: a way to opt files in/out of bundling might be a good idea. I can imagine wanting to bundle fundamental-to-the-design assets but not a video currently selected to play, for example.

vade:
thats odd - because i just exported fabric bundles and the asset paths were absolutely relative as i moved the document to the a new computer and it loaded. Are you certain you didnt open an old graph or version? 

The idea is the saved path should be relative, like this bundle document i just opened

image

I cant re-create this here - this is what i have when following those exact instructions:

image

> Also: a way to opt files in/out of bundling might be a good idea. I can imagine wanting to bundle fundamental-to-the-design assets but not a video currently selected to play, for example.

yea, right now the idea is that anything that does not have a patched output gets included (ie file paths that are not dynamically driven). There *might* be a way to have an intermediate UI / option panel that lets you check off available paths to be set or not - with the caveat shit breaks if you keep the path and move elsewhere...

tobyspark:
Aha. I must not have actually closed it – I’m multitasking! – so the actual issue is: saving a document does not re-write the paths in the open document. The only way to see them is to close and re-open.

vade:
if the document is unsaved, theres no relative path to make relative paths from (ie, the document does not exist on disk) so its sort of impossible to update until saved as a document bundle.

vade:
And, until next save of a fabric bundle, file paths are in fact absolute - as files are not dynamically copied into the bundle when added to the document - only on save are they copied into the bundle. 

Might that explain it?

tobyspark:
> anything that does not have a patched output gets included

Do you mean _input_? If so, that gets us the opt-out quite nicely – just a right-click on the port away _Insert Parameter Node_

vade:
ah, yes, sorry, i did in fact mean input - if the file path INPUT PORT is wired up - that file is skipped as part of asset copying. Sorry for the confusion!

tobyspark:
> And, until next save of a fabric bundle, file paths are in fact absolute - as files are not dynamically copied into the bundle when added to the document - only on save are they copied into the bundle.
> 
> Might that explain it?

I said it was still absolute _after save_. This is definitively the behaviour I see: only `asset` paths once the document has been re-opened.

vade:
Try the latest, should be resolved!

vade:
If that fixes it for ya, i'd love to merge in!

tobyspark:
Here’s a counter proposal. Two big reasons to consider it
1. It address more use-cases than bundling
2. Collating media and modifying documents is trickier than it looks, both technically and UX-wise

Proposal
- Fabric can resolve absolute and relative paths
- When a file path is created, if it shares a root with the document’s enclosing path it is written into the document as relative.
- On first save, i.e. when the document gains a location, paths in the document would be rewritten

That’s it. You get
- The bundle idiom by creating an assets folder next to the graph file.
- A project folder idiom by doing the above, and then duplicating your file into _v2_, or a bunch of sibling examples using the same media
- Zipping that folder for transport and archiving is arguably better than the bundle format

What you don’t get
- The per-file-format bundling decision not being what you want in a lot of cases, nomatter which per-document choice you make. We can’t treat e.g. a branding asset of tens of kb the same as a folder of movies coming to multi-GB. Doubly so if it results in a copy invisible to the user (where’s my disk space gone!). We could make some heuristics, but ultimately you’re going to need a UI to catalogue and manage asset bundling, with a bunch of “are you sure” dialogs in case the user isn’t looking at it.
- The failure modes around collapsing full paths into a filename-only set
- The failure modes around file operations
- The failure modes around those file operations being in sync with a live document

What’s a wash
- +ve Things like saving an updated asset over the existing file just work
- -ve The user can do things like that without realising it; they are managing their own folder

The work to do this is mostly this branch already, just with the bundle part taken out of it.

tobyspark:
From [discord conversation](https://discord.com/channels/1443372127427366933/1505751776366690445/1552717415081709629) – General agreement. Perhaps the value of bundles could be as export, which a) is a clear media management signal (yes, bundle everything!) and b) being disconnected from the live graph avoids a lot of the failure modes.

